PR #55, opened 2020-10-27
v10 of the draft

mjethanandani:
This is the base branch for all v10 changes. More commits will add to this branch.


jhaas-pfrc (commented):
We need to comment on how you read an as-path container in terms of indexing.  For example, consider path "1 2 3 [ 4 5 6 ]".  This would be two as-segments - one an as-sequence, one an as-set.  Having the sequence have an index of 1 and the set have an index of 2 is fine.  The leaf-list of 1,2,3 is also fine.

jhaas-pfrc (approved):

jhaas-pfrc (approved):

PR #56, opened 2020-10-30
Changes from part 2 of YANG doctors review

mjethanandani:


src/yang/ietf-bgp-common-structure.yang
```diff
@@ -76,7 +76,7 @@ submodule ietf-bgp-common-structure {
           "Configure logging of peer state changes.  Default is to
            enable logging of peer state changes.
 
-           Note: Documenting out of ESTABLISHED state is desirable,
+           Note: Documenting demotion of ESTABLISHED state is desirable,
```
jhaas-pfrc, line 79: demotion from ESTABLISHED

mjethanandani: Ok.

PR #58, opened 2020-11-15
Remove one more submodule

mjethanandani:
Fix to remove one more submodule, and updated the policy modules to reflect changes in ietf-routing-policy module. Also collapsed single usage of some of the groupings.


keyur-arrcus (approved):

src/yang/ietf-bgp-rib-tables.yang
```diff
@@ -70,6 +69,45 @@ submodule ietf-bgp-rib-tables {
...
+      type yang:timeticks;
+      description
+        "Timestamp when this path was last modified.
+
+         The value is the timestamp in seconds relative to
+         the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
+    }
+    leaf valid-route {
+      type boolean;
+      description
```
jhaas-pfrc, line 86: What exactly is "valid" meant to be referring to in the BGP spec sense of things?

src/yang/ietf-bgp-rib.yang
```diff
@@ -50,7 +46,9 @@ submodule ietf-bgp-rib {
 
      Authors: Mahesh Jethanandani (mjethanandani at gmail.com),
               Keyur Patel (keyur at arrcus.com),
-              Susan Hares (shares at ndzh.com)";
+              Susan Hares (shares at ndzh.com),
+              Jeffrey Haas (jhaas at pfrc.org).";
```
jhaas-pfrc, line 50: Could you update everything to be jhaas@juniper.net?  pfrc.org is just my workaround for dealing with ietf mailing lists when work's email is... not optimal.

src/yang/ietf-bgp-rib-tables.yang
mjethanandani: This might be completely Cisco thing, and Keyur can confirm. A route can be valid or invalid when the next-hop is unreachable. And then there is the whole discussion of what does it mean to validate a route in RFC 5123. But I am not an expert, so I will let the two of you tell me how you would rephrase this node, or remove it if it does not make sense.

src/yang/ietf-bgp-rib.yang
mjethanandani: Done

src/yang/ietf-bgp-rib-tables.yang
jhaas-pfrc: 5123 isn't a good match here.
If the intent is meant to be "is the route eligible for route install", then we can craft some words from rfc 4271.  
I'll wait on Keyur to respond.

keyur-arrcus (commented):
Route may not be valid (inbound policy rejects it) but it is still in table and flag is used to indicate the state?

src/yang/ietf-bgp-rib-tables.yang
jhaas-pfrc: This issue is still open.

jhaas-pfrc:
@keyur-arrcus 
> Route may not be valid (inbound policy rejects it) but it is still in table and flag is used to indicate the state?

If it's intended to signal that the route is ineligible for selection, the name would be better off changed to something that indicates that state.  valid/invalid implies something wrong with the route.

jhaas-pfrc:
> @keyur-arrcus
> 
> > Route may not be valid (inbound policy rejects it) but it is still in table and flag is used to indicate the state?
> 
> If it's intended to signal that the route is ineligible for selection, the name would be better off changed to something that indicates that state. valid/invalid implies something wrong with the route.

We'll address this in Issue #122

jhaas-pfrc (approved):

PR #99, opened 2021-06-20
Jhaas/issue 65

jhaas-pfrc:
BGP community matches should use the entire uint32 range while in uint32 mode.  Additionally, the pattern match should be restricted to match two colon-separated uint16 numbers.

The motivation to permit the full range even if an implementation doesn't like to be configured with them in some circumstances is that sometimes permitted match context may be different than permitted set context.  As an example, setting a community 0:* may be forbidden by an implementation, but may be permitted for matching purposes.


mjethanandani:
Adding Keyur to review the changes.

PR #102, opened 2021-06-20
Jhaas/issue 69

### LINKED ISSUE #69: cluster-id when clause wrong?
Consider the following below.

Shouldn't the when clause for route-reflector-client-id be when true rather than false?

Similarly, shouldn't the no-client-reflect share the same condition?

  grouping structure-neighbor-group-route-reflector {
    description
      "Structural grouping used to include route reflector
       configuration and state for both BGP neighbors and peer
       groups";
    container route-reflector {
      description
        "Route reflector parameters for the BGP peer-group";
      reference
        "RFC 4456: BGP Route Reflection.";
      leaf route-reflector-cluster-id {
        when "../route-reflector-client = 'false'";
        type bt:rr-cluster-id-type;
        description
          "Route Reflector cluster id to use when local router is
           configured as a route reflector.  Commonly set at the
           group level, but allows a different cluster id to be set
           for each neighbor.";
        reference
          "RFC 4456: BGP Route Reflection: An Alternative to
                     Full Mesh.";
      }
      leaf no-client-reflect {
        type boolean;
        default "false";
        description
          "When set to 'true', this disables route redistribution
           by the Route Reflector. It is set 'true' when the client is
           fully meshed to prevent sending of redundant route
           advertisements.";
      }
      leaf route-reflector-client {
        type boolean;
        default "false";
        description
          "Configure the neighbor as a route reflector client.";
        reference
          "RFC 4456: BGP Route Reflection: An Alternative to
                     Full Mesh.";
      }
    }
  }

jhaas-pfrc:
When condition on route-reflector-client was incorrect.  Also clarify description for no-client-reflect.

Closes #69

### DIFF
--- a/src/yang/ietf-bgp-common-structure.yang
+++ b/src/yang/ietf-bgp-common-structure.yang
@@ -120,7 +120,7 @@ submodule ietf-bgp-common-structure {
       reference
         "RFC 4456: BGP Route Reflection.";
       leaf route-reflector-cluster-id {
-        when "../route-reflector-client = 'false'";
+        when "../route-reflector-client = 'true'";
         type bt:rr-cluster-id-type;
         description
           "Route Reflector cluster id to use when local router is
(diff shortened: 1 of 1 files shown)


src/yang/ietf-bgp-common-structure.yang
```diff
@@ -132,13 +132,14 @@ submodule ietf-bgp-common-structure {
...
       leaf no-client-reflect {
+        when "../route-reflector-client = 'true'";
         type boolean;
         default "false";
         description
           "When set to 'true', this disables route redistribution
            by the Route Reflector. It is set 'true' when the client is
-           fully meshed to prevent sending of redundant route
-           advertisements.";
+           fully meshed in its peer-group to prevent sending of redundant
```
mjethanandani, line 141: Can you check to make sure the column width does not exceed 70.

jhaas-pfrc:
After discussion with Mahesh and Sue, the fix is wrong.  Closing to try again.

PR #109, opened 2021-06-20
Jhaas/issue 72

### LINKED ISSUE #72: Keepalive configuration default problematic?
Consider:
    leaf keepalive {
      type uint16 {
        range "0..21845";
      }
      units "seconds";
      default "30";

The behavior in a number of implementations is to only have a holdtime configurable.  In such cases, if you configure the holdtime, the keepalive is automatically derived as 1/3 of holdtime.

I believe that in the presence of the default above, there's not a good way to enable that behavior.  The absence of configuration for keepalive in the presence of default implies 30.

I'd recommend removing the default, and noting the behavior we want in the description.

jhaas-pfrc:
Remove default for keepalive. It's either explicitly configured or picks up its operational default via the description.

Closes #72

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -113,7 +113,6 @@ submodule ietf-bgp-common {
         range "0..21845";
       }
       units "seconds";
-      default "30";
       description
         "When used as a configuration (rw) value, this Time interval
          (in seconds) for the KeepAlive timer configured for this BGP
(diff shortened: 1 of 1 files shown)


mjethanandani:
I do not know if I agree with the statement that the default value for keepalive will be picked up from the description statement. The description statement **expects** an implementation to actually set that value when the connection is set up. Having a default value **forces** an implementation to either accept the value as is or change it if they disagree. Two very different actions.

suehares:
I believe that we should set the default to 30 seconds, but provide RFC 4271 reference to section 10.

jhaas-pfrc:
Closing this pull request in favor of different text.

PR #110, opened 2021-06-20
Jhaas/issues 78

### LINKED ISSUE #78: indexes are under specified for as-paths
We have an as-path segment index, as noted below.

We do need an index to support more than one segment.  However, the index doesn't specify what we do here. 

Is the index a continuous range?  E.g. 1,2,3?  Is it zero based or one-based?  How do we ordering of the segments - is it left to right as in the AS-path and are they numbered 1,2,3 or 3,2,1?

  grouping bgp-as-path-attr {                                                                                
    description                                                                                              
      "Data for representing BGP AS-PATH attribute";                                                         
    leaf index {                                                                                             
      type uint16;                                                                                           
      description                                                                                            
        "An index into the AS segments.";                                                                    
    }

jhaas-pfrc:
Remove index from bgp-as-path-attr.

Closes #78

### DIFF
--- a/src/yang/ietf-bgp-rib-attributes.yang
+++ b/src/yang/ietf-bgp-rib-attributes.yang
@@ -60,11 +60,6 @@ submodule ietf-bgp-rib-attributes {
   grouping bgp-as-path-attr {
     description
       "Data for representing BGP AS-PATH attribute";
-    leaf index {
-      type uint16;
-      description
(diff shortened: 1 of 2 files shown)


mjethanandani (approved):

keyur-arrcus (commented):
Looks good to me.

PR #116, opened 2021-06-22
Jhaas/issue 62

### LINKED ISSUE #62: Should route-selection-options be per-group/neighbor or just global?
grouping route-selection-options is applied at multiple scopes.

Some of these options are reasonable per-peer:
- MED
- enable-aigp
- advertise-inactive

Some of these are likely intended to be consistent for the entire BGP implementation:
- always-compare-med
- ignore-as-path-length
- external-compare-router-id
- ignore-nexthop-igp-metric

The med feature governs both sending of MEDs and utliizing it, which might need to be distinct properties.

The general distinction of these two things is there's a cluster of items governing things we put into the protocol, while other govern how we handle them on receive for route-selection purposes.

I suspect these should be split.  I also suspect that the second cluster is likely only global-scoped.

jhaas-pfrc:
Issue 62: route-selection-options are not for neighbor configuration

Closes #62

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -458,7 +458,6 @@ module ietf-bgp {
                RFC 7454: BGP Operations and Security.";
           }
           uses neighbor-group-config;
-          uses route-selection-options;
           leaf session-state {
             type enumeration {
               enum idle {


mjethanandani (approved):

src/yang/ietf-bgp.yang
```diff
@@ -458,7 +458,6 @@ module ietf-bgp {
                RFC 7454: BGP Operations and Security.";
           }
           uses neighbor-group-config;
-          uses route-selection-options;
```
keyur-arrcus, line 461: looks ok to me.

keyur-arrcus (approved):

PR #117, opened 2021-06-23
Address issue#76

mjethanandani:
Most of the changes are related to getting the draft to "compile" with the latest version of xml2rfc and to deal with changes in models that this model depends on, e.g. ietf-routing-policy.

The remaining changes are tooling changes, and they relate mostly to the strict changes that yanglint has imposed.


jhaas-pfrc (approved):
Looks fine.

PR #127, opened 2021-06-28
Issue 85: clear bgp needs additional arguments

jhaas-pfrc:
Issue 85: clear bgp needs additional arguments

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -858,6 +858,54 @@ module ietf-bgp {
             "Clear neighbors action.";
 
           input {
+	    choice operation {
+	      default operation-admin;
+	      description
+	        "The type of operation for the clear action.";
+	      case operation-admin {
+	        leaf admin {
+		  type empty;
+		  description
+		    "Closes the Established BGP session with a BGP
+		     NOTIFICATION message with the Administrative
+		     Reset error subcode.";
+		  reference
+		    "RFC 4486.";
+		}
+	      }
+	      case operation-hard {
+	        leaf hard {
+		  type empty;
+		  description
+		    "Closes the Established BGP session with a BGP
+		     NOTIFICATION message with the Hard Reset error
+		     subcode.";
+		  reference
+		    "RFC 8538, Section 3.";
+		}
+	      }
+	      case operation-soft {
+	        leaf soft {
+		  type empty;
+		  description
+		    "Re-sends the current Adj-Rib-Out to this
+		     neighbor.";
+		}
+	      }
+	      case operation-soft-inbound {
+	        leaf soft-inbound {
+		  // XXX JMH - if we wanted to permit future expansion
+		  // to be per-AFI/SAFI, is this a correct definition?
+		  if-feature "bt:route-refresh";
+		  type empty;
+		  description
+		    "Requests the Adj-Rib-In for this neighbor to be
+		     re-sent using the BGP Route Refresh feature.";
+		}
+	      }
+	    }
+
+
             leaf clear-at {
               type yang:date-and-time;
               description
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -858,52 +858,54 @@ module ietf-bgp {
             "Clear neighbors action.";
 
           input {
-	    choice operation {
-	      default operation-admin;
-	      description
-	        "The type of operation for the clear action.";
-	      case operation-admin {
-	        leaf admin {
-		  type empty;
-		  description
-		    "Closes the Established BGP session with a BGP
-		     NOTIFICATION message with the Administrative
-		     Reset error subcode.";
-		  reference
-		    "RFC 4486.";
-		}
-	      }
-	      case operation-hard {
-	        leaf hard {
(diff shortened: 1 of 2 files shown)


jhaas-pfrc:
Note there is a question in the soft-inbound case that needs to be resolved.  Any suggestions @mjethanandani ?

src/yang/ietf-bgp.yang
```diff
@@ -858,6 +858,54 @@ module ietf-bgp {
...
+	        "The type of operation for the clear action.";
+	      case operation-admin {
+	        leaf admin {
+		  type empty;
+		  description
+		    "Closes the Established BGP session with a BGP
+		     NOTIFICATION message with the Administrative
+		     Reset error subcode.";
+		  reference
+		    "RFC 4486.";
```
mjethanandani, line 873: In addition to citing the RFC number, I usually provide the title of the RFC. The format for the reference is:

"RFC XXXX: <Title of the RFC>, <Optionally, the word Section followed by the Section Number>.";

src/yang/ietf-bgp.yang
```diff
@@ -858,6 +858,54 @@ module ietf-bgp {
             "Clear neighbors action.";
 
           input {
+	    choice operation {
```
mjethanandani, line 861: Not sure if this is indented correctly.

src/yang/ietf-bgp.yang
```diff
@@ -858,6 +858,54 @@ module ietf-bgp {
...
+	      }
+	      case operation-hard {
+	        leaf hard {
+		  type empty;
+		  description
+		    "Closes the Established BGP session with a BGP
+		     NOTIFICATION message with the Hard Reset error
+		     subcode.";
+		  reference
+		    "RFC 8538, Section 3.";
```
mjethanandani, line 884: Add the full reference here also.

src/yang/ietf-bgp.yang
```diff
@@ -858,6 +858,54 @@ module ietf-bgp {
...
+		}
+	      }
+	      case operation-soft {
+	        leaf soft {
+		  type empty;
+		  description
+		    "Re-sends the current Adj-Rib-Out to this
+		     neighbor.";
+		}
```
mjethanandani, line 893: Any reference for these operations?

src/yang/ietf-bgp.yang
```diff
@@ -858,6 +858,54 @@ module ietf-bgp {
...
+	      }
+	      case operation-soft-inbound {
+	        leaf soft-inbound {
+		  // XXX JMH - if we wanted to permit future expansion
+		  // to be per-AFI/SAFI, is this a correct definition?
```
mjethanandani, line 898: I am not clear on what the description of leaf 'soft-inbound' is? Is of type 'afi-safi-type' or 'empty'?

Also is the entire case of 'operation-soft-inbound' controlled by the feature statement, or just the setting of 'soft-inbound'?

mjethanandani (changes requested):

src/yang/ietf-bgp.yang
jhaas-pfrc: I've added the RFC title.  No section is added since the entire RFC is about those codes.

src/yang/ietf-bgp.yang
jhaas-pfrc: The section had tabs.  Turns out my .vimrc wasn't being properly loaded.  I believe I've now fixed that.

src/yang/ietf-bgp.yang
jhaas-pfrc: I've done so, but we're inconsistently documenting references.  4271 is cited everywhere for obvious reasons, but without similar expansion.

src/yang/ietf-bgp.yang
mjethanandani: Ahh! The confusion was around the XXX comment.

And now that I understand the issue, the augmentation would have to be to the case statement to add another leaf that indicates the afi/safi type in addition to the existing leaf. 

HTH.

src/yang/ietf-bgp.yang
```diff
@@ -858,6 +858,56 @@ module ietf-bgp {
...
+                leaf soft {
+                  type empty;
+                  description
+                    "Re-sends the current Adj-Rib-Out to this
+                     neighbor.";
+                }
+              }
+              case operation-soft-inbound {
+                leaf soft-inbound {
+                  // XXX JMH - if we wanted to permit future expansion
```
mjethanandani, line 899: Please remove XXX comment before committing the change, if you are happy with the explanation.

mjethanandani (approved):

src/yang/ietf-bgp.yang
suehares: Ok with RFC

suehares:
Ok with jeff's comments.

src/yang/ietf-bgp.yang
jhaas-pfrc: Okay, additional leaves with the presence node in the same choice is a reasonable way this can be extended.  I'll remove the XXX and commit.

Issue #86, opened 2021-06-17
BFD configuration


jhaas-pfrc:
Some implementations of BGP has their BFD configuration as an interface specific configuration.
Some, e.g. JunOS, has its BFD configuration as per-neighbor.

we should figure out how to be supportive of both.

suehares:
+1 on supporting BFD per-neighbor.

mjethanandani:
The BGP model only enables/disables an interface for BFD. The part of configuring which neighbor receives is part of BFD configuration, as far as I can see:

       +--rw ip-sh
          +--rw sessions
          |  +--rw session* [interface dest-addr]
          |     +--rw interface                         if:interface-ref
          |     +--rw dest-addr                         inet:ip-address
          |     +--rw source-addr?                      inet:ip-address
          |     +--rw local-multiplier?                 multiplier

jhaas-pfrc:
I'll be proposing a patch maybe next week.

PR #137, opened 2021-10-21
Add BFD config at a per neighbor level

mjethanandani:
The current model supports BFD configuration at an interface level. However, some implementations allow for BFD configuration at per neighbor level, e.g. Juniper. This change adds that ability.

Also updated an example to demonstrate that configuration.

The rest of the changes were made more to get the draft to compile (again) with yanglint (which has upped its error checks).

Approving this CR should close issue #86.

### DIFF
--- a/src/yang/ietf-bgp-commands.yang
+++ b/src/yang/ietf-bgp-commands.yang
@@ -33,6 +33,7 @@ module ietf-bgp-commands {
     uses bgp:attr-set-attributes;
     uses bgp:bgp-community-attr-state;
     uses bgp:ext-community-attributes;
+    uses bgp:large-community-attributes;
     uses bgp:bgp-common-route-annotations-state;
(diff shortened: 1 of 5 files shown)


keyur-arrcus (approved):
Looks good under neighbors

src/yang/ietf-bgp.yang
```diff
@@ -956,7 +956,10 @@ module ietf-bgp {
                  interface.";
             }
             description
-              "BFD configuration parameters.";
+              "BFD client configuration.";
+            reference
+              "I-D.ietf-bfd-rfc9127-bis: YANG Data Model for
+	      Bidirectional Forward Detection (BFD).";
```
jhaas-pfrc, line 962: Indentation issue.

mjethanandani: Will fix.

jhaas-pfrc (approved):

Issue #87, opened 2021-06-17
clear routes isn't clear?


jhaas-pfrc:
What feature is clear routes intended to cover?  The documentation doesn't have enough detail.

  grouping clear-routes {
    description
      "Action to clear BGP routes.";
    container clear-routes {
      if-feature "bt:clear-routes";
      action clear {

suehares:
Is this clearing AS path and leaving prefix in-active or is it removing the entire rout.

mjethanandani:
At least one implementation does a 'clear ip bgp route *', and it clears rib-in.

PR #139, opened 2021-10-27
Clarify clear-route action

mjethanandani:
This PR addresses item #87. It adds a description statement under every use of the 'clear-route' grouping as to what it is clearing.

What was discussed on the call did not include what happens with ipv* adj-rib-in-post. Added a 'clear-route' option for that also.

Finally, refactored the grouping ipv*-loc-rib and ipv*-adj-rib grouping out as it was being used only once.


keyur-arrcus (approved):
Looks good as well. Although I believe we will need this grouping in future.

PR #144, opened 2021-12-27
Issue 86: Per-client bfd configuration

jhaas-pfrc:
After the discussion that resulted in RFC9127-bis in BFD, the fix here
is to simply move the grouping to the transport parameters.

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -14,6 +14,12 @@ submodule ietf-bgp-common {
     reference
       "RFC 6991: Common YANG Data Types.";
   }
+  import ietf-bfd-types {
+    prefix bfd-types;
+    reference
+      "RFC 9127, YANG Data Model for Bidirectional Forward
+       Detection.";
+  }
 
   organization
     "IETF IDR Working Group";
@@ -312,7 +318,6 @@ submodule ietf-bgp-common {
          expressed as either an IP address or reference to the name
          of an interface.";
     }
-
     leaf md5-auth-password {
       type string;
       description
@@ -321,6 +326,15 @@ submodule ietf-bgp-common {
       reference
         "RFC 2385: Protection of BGP Sessions via the TCP MD5
                    Signature Option.";
+    }
+    container bfd {
+      if-feature "bt:bfd";
+      uses bfd-types:client-cfg-parms;
+      description
+        "BFD client configuration.";
+      reference
+        "RFC 9127, YANG Data Model for Bidirectional Forwarding
+         Detection.";
     }
   }
 
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -17,7 +17,7 @@ submodule ietf-bgp-common {
   import ietf-bfd-types {
     prefix bfd-types;
     reference
-      "RFC 9127, YANG Data Model for Bidirectional Forward
+      "RFC XXXX, YANG Data Model for Bidirectional Forward
        Detection.";
   }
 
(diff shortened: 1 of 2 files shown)


jhaas-pfrc:
Note that this would replace pull request 137.

src/yang/ietf-bgp-common.yang
```diff
@@ -14,6 +14,12 @@ submodule ietf-bgp-common {
     reference
       "RFC 6991: Common YANG Data Types.";
   }
+  import ietf-bfd-types {
+    prefix bfd-types;
+    reference
+      "RFC 9127, YANG Data Model for Bidirectional Forward
```
mjethanandani, line 20: This will have to reference rfc9127-bis.

src/yang/ietf-bgp-common.yang
```diff
@@ -322,6 +327,15 @@ submodule ietf-bgp-common {
         "RFC 2385: Protection of BGP Sessions via the TCP MD5
                    Signature Option.";
     }
+    container bfd {
```
mjethanandani, line 330: Are you proposing to make the bfd container part of the neighbor-group-transport-config grouping? If so, and just to let you know, that grouping is used not only in the main BGP module neighbor config, but also in the peer group configuration. If that is not intended, why not keep the container under the neighbor config in the main BGP module, like it was before.

src/yang/ietf-bgp-common.yang
```diff
@@ -322,6 +327,15 @@ submodule ietf-bgp-common {
...
+      if-feature "bt:bfd";
+      uses bfd-types:client-cfg-parms;
+      description
+        "BFD client configuration.";
+      reference
+        "RFC 9127, YANG Data Model for Bidirectional Forwarding
```
mjethanandani, line 336: This reference should also be updated to point to 9127-bis.

src/yang/ietf-bgp-common.yang
jhaas-pfrc: I'll make this XXXX for now and open a new issue for updating it after we're through the 9127-bis dance.

src/yang/ietf-bgp-common.yang
jhaas-pfrc: I'll make this XXXX for now and open a new issue for updating it after we're through the 9127-bis dance.

src/yang/ietf-bgp-common.yang
jhaas-pfrc: BFD configuration should be part of group configuration.  As part of trying to understand if that was an issue, I opened issue 154 that notes other discrepancies like the security settings.

While it's possible for security mechanisms to need to be per-peer, in many cases they are per-group and selectively overridden if it is necessary.

mjethanandani: So you are suggesting that BFD configuration applies to peer group configuration also. How would that work? I mean, if I have configured BFD with some parameters at the local end, how am I conveying them to the peer? Maybe I am missing something here.

jhaas-pfrc: > So you are suggesting that BFD configuration applies to peer group configuration also. How would that work? I mean, if I have configured BFD with some parameters at the local end, how am I conveying them to the peer? Maybe I am missing something here.

Remember, this is the client's desired usage for BFD.  A peer-group is just a template that is shared by a number of neighbors.  Thus, saying "yes, I want bfd enabled, local-multiplier 3, 10ms interval" for the group is appropriate.

A peer group doesn't instantiate peering.  A neighbor must be configured that uses the peer-group. See /bgp/neighbors[ip]/peer-group leaf.

mjethanandani (approved):

keyur-arrcus (approved):
looks good.

PR #150, opened 2021-12-27
Issue 60: Restructure graceful restart to properly reflect forwarding…

### LINKED ISSUE #60: remote-helper
Acee (as a YANG doctor) commented that:

    5. Page 29, "remote-helper": I'd support this to be renamed
       "graceful-restart-only" since "helper-only" is relative to the local
         peer.

To which I responded:
 
The enumeration definitions talk about the mode in terms of local and remote. Calling it ‘graceful-restart-only’ conveys a completely different meaning.

Acee came back with:

I think this is a poor choice since it should be terms of the device – not the remote device. Perhaps, using your line of thought, even “local-graceful-restart” would be much better.

Jeff chimed in to say:

There's actually a few issues here:
- The restarting mode is per AFI-SAFI, while this container doesn't contain
 anything that contains per AFI-SAFI state.
- The local peer can know whether it's acting as a helper and whether it can
 retain forwarding for a given AFI/SAFI.
- We can know if the remote BGP speaker has advertised GR for a given
 AFI/SAFI, which tells us whether it supports GR and can help the local
 speaker with restarting.
- We can know if, on the last restart, forwarding was preserved for that
 AFI/SAFI.

Proposal: Replace leaf "mode" with:
list afi-safi-supported {
 key "afi-safi-name"
 leaf afi-safi-name {
   type identityref { base bt:afi-safi-type; }
 }
 leaf local-supported {
   type boolean;
 }
 leaf remote-supported {
   config false;
   type boolean;
 } 
 leaf remote-forweard-preserved {
   type boolean;
   config false;
 }
}

Acee said:

You need some good descriptions for the booleans as the names aren't enough. I'd replace "remote-forward-preserved" with "remote-forwarding-preserved" as in Forwarding Information Base (FIB).  Note that you no longer have the granularity between support of local restart and local helper (based on only one config true leaf).  Also, it seems that granularity is now wrong - you need the config false values on a per-peer basis so I don't think you can mix the local-configuration with the remote peer state in the same grouping. 

FWIW here is the OSPF config true container.

container graceful-restart {
     if-feature graceful-restart;
     description
       "Graceful restart config state.";
     reference "RFC 3623: OSPF Graceful Restart                                                                                                                         
                RFC 5187: OSPFv3 Graceful Restart";                                                                                                                         
     leaf enable {
       type boolean;
       description
         "Enable/Disable graceful restart as defined in RFC 3623                                                                                                    
          for OSPFv2 and RFC 5187 for OSPFv3.";                                                                                                                          
     }
     leaf helper-enable {
       type boolean;
       description
         "Enable graceful restart helper support for restarting                                                                                                           
          routers (RFC 3623 Section 3).";                                                                                                                                          
     }
     leaf restart-interval {
       type uint16 {
          range "1..1800";
       }
       units seconds;
       default "120";
       description
         "Interval to attempt graceful restart prior                                                                                                                              
          to failing (RFC 3623 Section B.1) (seconds)";                                                                                                                     
     }
     leaf helper-strict-lsa-checking {
       type boolean;
       description
         "Terminate graceful restart when an LSA topology change                                                                                                  
          is detected (RFC 3623 Section B.2).";                                                                                                                                 
     }
   }

To Jeff's suggestion I asked:

I am fine with making the change. However, are the three leafs, local-supported, remote-supported, remote-forward-preserved mutually exclusive per AFI/SAFI, or can the local BGP speaker be configured for any two, or all three possibilities? In other words, should the “support” leafs be a choice statement?

Keyur commented:

This makes sense.

What is the consensus?

jhaas-pfrc:
… state

Acee Lindem flagged during YANG doctor review that the "mode" leaf
didn't quite cover proper forwarding state preservation semantics.
Related IDR list followup discussed how this state also needed to be
per-AFI-SAFI.

The fix was primarily to remove it from global scope and make it
per-neighbor, per AFI-SAFI state.

While I was in there, I also added a flag to indicate whether End-of-RIB
was received from the peer for that AFI-SAFI.  This is a core bit of
useful troubleshooting infrastructure most implementations have for the
graceful restart feature.

Closes #60

### DIFF
--- a/src/yang/ietf-bgp-neighbor.yang
+++ b/src/yang/ietf-bgp-neighbor.yang
@@ -190,6 +190,35 @@ submodule ietf-bgp-neighbor {
             "This leaf indicates whether the ability to support
              graceful-restart has been advertised to the peer";
         }
+        leaf local-forwarding-state-preserved {
+          type boolean;
+          config false;
+          description
+            "This leaf indicates whether the local router has
+             or would advertise the Forwarding State bit in its
+             Graceful Restart capability for this AFI-SAFI.";
+          reference
+            "RFC 4724: Graceful Restart Mechanism for BGP.";
+        }
+        leaf forwarding-state-preserved {
+          type boolean;
+          config false;
+          description
+            "This leaf indicates whether the neighbor has advertised
+             the Forwarding State bit in its Graceful Restart
+             capability for this AFI-SAFI.";
+          reference
+            "RFC 4724: Graceful Restart Mechanism for BGP.";
+        }
+        leaf end-of-rib-received {
+          type boolean;
+          config false;
+          description
+            "This leaf indicates whether the neighbor has advertised
+             the End-of-RIB marker for this AFI-SAFI.";
+          reference
+            "RFC 4724: Graceful Restart Mechanism for BGP.";
+        }
       }
       uses mp-all-afi-safi-list-contents;
       uses bgp-neighbor-use-multiple-paths;
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -615,38 +615,10 @@ module ietf-bgp {
               type boolean;
               config false;
               description
-                "This flag indicates whether the local neighbor is
+                "This flag indicates whether this BGP session is
                  currently restarting. The flag is cleared after all
                  NLRI have been advertised to the peer, and the
                  End-of-RIB (EOR) marker has been cleared.";
-            }
-            leaf mode {
-              type enumeration {
-                enum helper-only {
-                  description
-                    "The local router is operating in helper-only
-                     mode, and hence will not retain forwarding state
-                     during a local session restart, but will do so
-                     during a restart of the remote peer";
-                }
-                enum bilateral {
-                  description
-                    "The local router is operating in both helper
-                     mode, and hence retains forwarding state during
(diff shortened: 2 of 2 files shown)


src/yang/ietf-bgp-neighbor.yang
```diff
@@ -190,6 +190,35 @@ submodule ietf-bgp-neighbor {
...
+          type boolean;
+          config false;
+          description
+            "This leaf indicates whether the local router has
+             or would advertise the Forwarding State bit in its
+             Graceful Restart capability for this AFI-SAFI.";
+          reference
+            "RFC 4724: Graceful Restart Mechanism for BGP.";
+        }
+        leaf forwarding-state-preserved {
```
keyur-arrcus, line 203: Can you Rename this as a neighbor-forwarding-state-preserved or something that indicates its for the neighbor state?

keyur-arrcus (approved):
Approving with a comment so as to reduce the churn.

src/yang/ietf-bgp-neighbor.yang
jhaas-pfrc: The neighbor portion is implied by the scope.  Do we really need it in there?

keyur-arrcus: Ack. nah. I see what you mean.

PR #151, opened 2021-12-27
Issue 131: Model RFC 2545 link-local nexthops.

### LINKED ISSUE #131: Modeling RFC 2545 double-nexthop?
We support BGP nexthop.  However, we don't have a rfc 2545 double nexthop option.

Update this issue to discuss whether we carry link local as well.

jhaas-pfrc:
Closes #131

### DIFF
--- a/src/yang/ietf-bgp-rib.yang
+++ b/src/yang/ietf-bgp-rib.yang
@@ -142,6 +142,16 @@ submodule ietf-bgp-rib {
         reference
           "RFC 4271: Section 5.1.3.";
       }
+      leaf link-local-next-hop {
+        type inet:ipv6-address;
+        description
+          "When both a global and a link-local next-hop are sent
+           when following RFC 2545 procedures, this leaf contains
+           the link-local next-hop.";
+        reference
+          "RFC 2545: Use of BGP-4 Multiprotocol Extensions for IPv6
+           Inter-Domain Routing";
+      }
       leaf med {
         type uint32;
         description


src/yang/ietf-bgp-rib.yang
```diff
@@ -142,6 +142,16 @@ submodule ietf-bgp-rib {
         reference
           "RFC 4271: Section 5.1.3.";
       }
+      leaf link-local-next-hop {
```
keyur-arrcus, line 145: This is under the common grouping of attr-set-attributes. How do we control it for safis where this is not relevant? we need a conditional check?

jhaas-pfrc: Our choices are:
1. It's an optional leaf, so it simply drops out when not in use.
2. We can add a conditional clause that limits it for ipv6 address families.  However, that starts to entangle maintenance of this leaf based on address families that may not yet be at this point.  Recall that our definition is an identity that is the tuple.

I'm up for better ideas.

suehares: Why is an optional leaf not a good choice?

jhaas-pfrc: In further review, I believe these leaves are only available in a read-only context.  If this was a read-write context for config state, it might be more problematic.

keyur-arrcus: Ok with any of the option suggested. Thinking a bit more, we can also move this as a backend validation check (along with or without option 1 or 2).

jhaas-pfrc: I think we've converged on this point.  The answer is we just leave it and it only is used when required. 
If so, please approve review.

keyur-arrcus (approved):

suehares (approved):

suehares (approved):

PR #152, opened 2021-12-27
Fix Docker build

mjethanandani:


src/validate-and-gen-trees.sh
```diff
@@ -4,7 +4,8 @@
 # Does the user have all the IETF published models.
 #
 if [ ! -d ../../yang-parameters ]; then
-   rsync -avz --delete rsync.iana.org::assignments/yang-parameters ../../
+    rsync -avz --delete rsync.iana.org::assignments/yang-parameters ../../
+    rm -rf ../../yang-parameters/*bfd*
```
jhaas-pfrc, line 8: This will collide with my issue.  Did you want me to hold off committing that?

src/yang/ietf-bgp.yang
```diff
@@ -851,9 +851,12 @@ module ietf-bgp {
               "RFC 4271, Section 8.2.2.";
           }
           description
-            "The backward-transition event is
+            "The backward-transition event is generally
```
jhaas-pfrc, line 854: Don't include this in the fix build commit.

src/validate-and-gen-trees.sh
mjethanandani: I will resolve this before merging.

src/yang/ietf-bgp.yang
mjethanandani: I believe this came to me as part of merge I did with v13.

mjethanandani: On re-checking, apparently not. Will remove it.

jhaas-pfrc (approved):

Issue #153, opened 2021-12-28
bgp neighbor leaf last-error is config true?


jhaas-pfrc:
This is noted in the file and shown in the yang tree.  It's unclear whether the issue here was a yang validator issue or not.

If this is solely an issue with the validator, we should get that addressed and make this node config false, which is appropriate.

          leaf last-error {
            type binary {
              length "2";
            }
            // notification does not like non-config statement.
            // config false;

jhaas-pfrc:
I tried restoring config false to this leaf.  I believe things are validating properly.  Is there reason to leave things in the current state?

PR #157, opened 2021-12-28
Issue #153. Fix the config false statement in last-error

mjethanandani:
Issue #153

### DIFF
--- a/draft/draft-ietf-idr-bgp-model.xml
+++ b/draft/draft-ietf-idr-bgp-model.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="US-ASCII"?>
 <!DOCTYPE rfc SYSTEM "rfc2629.dtd">
-<rfc category="std" docName="draft-ietf-idr-bgp-model-latest" ipr="trust200902" consensus="true">
+<rfc category="std" docName="draft-ietf-idr-bgp-model-latest" ipr="trust200902">
   <?rfc toc="yes"?>
 
   <front>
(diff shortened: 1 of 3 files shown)


draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="US-ASCII"?>
 <!DOCTYPE rfc SYSTEM "rfc2629.dtd">
-<rfc category="std" docName="draft-ietf-idr-bgp-model-latest" ipr="trust200902">
+<rfc category="std" docName="draft-ietf-idr-bgp-model-latest" ipr="trust200902" consensus="true">
```
jhaas-pfrc, line 3: I suggest that changes to the code outside the scope of the issue get fixed under their own issue.

src/validate-and-gen-trees.sh
```diff
@@ -98,7 +98,7 @@ rm ../bin/*-rib-tree.txt.tmp
 
 echo "Validating examples"
 
-for i in yang/example-bgp-configuration-*.xml
+for i in yang/example-bgp-configuration-a.*.xml
```
jhaas-pfrc, line 101: I suggest that changes to the code outside the scope of the issue get fixed under their own issue.

mjethanandani: I would love to, and understand the sentiment behind keeping the commit limited to the immediate issue at hand.

Except without this change, I am getting compile errors and the draft does not build and to fix all of that requires me to do even more surgery than I would like to.

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: I can remove this.

src/validate-and-gen-trees.sh
jhaas-pfrc: Understood.  It wasn't an obvious place where the change would result in errors.  I did try it out and don't think I noticed anything.  I must have missed it.

jhaas-pfrc (approved):

PR #158, opened 2021-12-30
Discrepancies between peer-group and neighbor configuration elements

mjethanandani:
Have tried to reorganize the two containers with similar content to make sure they are in similar order accounting for extra/different elements.

Please take a close look at the tree diagram, as I still see discrepancies between the two containers, e.g. graceful-restart, and was not sure how much to align the parameters within the container.

[ietf-bgp@2021-12-30-tree.txt](https://github.com/mjethanandani/ietf-bgp-yang/files/7794136/ietf-bgp%402021-12-30-tree.txt)


src/yang/ietf-bgp.yang
```diff
@@ -281,11 +278,16 @@ module ietf-bgp {
 
         list neighbor {
           key "remote-address";
-
           description
             "List of BGP neighbors configured on the local system,
              uniquely identified by remote IPv[46] address.";
 
+          leaf remote-address {
```
jhaas-pfrc, line 285: What's the motivation to move this up?  I realize that it's the key.  The motivation for the current ordering is tcp sessions are typically pictured in the sense of local:remote.

src/yang/ietf-bgp.yang
```diff
@@ -924,11 +928,156 @@ module ietf-bgp {
           }
         }
       }
+
       container peer-groups {
         description
           "Configuration for BGP peer-groups";
-        uses bgp-peer-group-list;
```
jhaas-pfrc, line 930: I understand your intent here.  I'd like some time to stare at the contents to see if we can try to address the issue I opened to know if we can still leverage a common grouping to avoid duplication of code.

src/yang/ietf-bgp.yang
mjethanandani: The motivation is because of the fact that just as 'name' is the key for neighbor and appears first, 'remote-address' is key for peer-group and appears first. Also note that peer-group has plenty of other parameters that are not present in neighbor, so the order is not going to be perfect.

src/yang/ietf-bgp.yang
mjethanandani: That is entirely possible, and I can see myself defining some of the items as groupings. But let us get the sets to be common before I create them.

src/yang/ietf-bgp.yang
keyur-arrcus: This comment says other way around. Neighbor is indexed by IP address and peer-group is indexed by a name.

keyur-arrcus (approved):
This grouping looks good to me. Happy to chat on the call if needed.

src/yang/ietf-bgp.yang
mjethanandani: Updated the comment.

src/yang/ietf-bgp.yang
mjethanandani: Created a common grouping to avoid duplication of code.

PR #161, opened 2021-12-31
Issue-145: Make send-community use an identity rather than bits for long term extensibility

### LINKED ISSUE #145: Should community-type bit positions be fully enumerated?
Consider the community-type typedef below.

The rules for "position" in RFC 7950 basically say that each subsequent bit gets the +1 assignment.

However, consider extensions that may come in the future.  Given such extensions will likely result in some implementations with new community-like feature that may cross as ships in the night, there's no guarantee that new-community-feature1 will be directly comparable to new-community-feature2.  This may result in config generated against schemas of potentially different versions to generate the wrong bit value.

If my interpretation is correct, I suspect this means two things:
1. We should fully enumerate bit positions in our current file.
2. We should recommend that such positions are IANA coordinated in a registry.  Is there a procedure for that?

  typedef community-type {                                                                                       
    type bits {
      bit standard {                                                                                             
        position 0;                                                                                              
        description
          "Send standard communities.";                                                                          
        reference
          "RFC 1997: BGP Communities Attribute.";                                                                
      }
      bit extended {                                                                                             
        description
          "Send extended communities.";                                                                          
        reference
          "RFC 4360: BGP Extended Communities Attribute.";                                                       
      }
      bit large {
        description
          "Send large communities.";                                                                             
        reference
          "RFC 8092: BGP Large Communities Attribute.";                                                          
      }                                                                                                          
    }
    description
      "Type describing variations of community attributes.
       The community types can be combined and a value of 0                                                      
       implies 'none'";                                                                                          
  }

jhaas-pfrc:
Closes #145

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -258,13 +258,14 @@ submodule ietf-bgp-common {
         "Routes learned via BGP are subject to weighted route
          dampening.";
     }
-    leaf send-community {
+    leaf-list send-community {
       if-feature "bt:send-communities";
-      type bt:community-type;
+      type identityref {
+        base "bt:send-community-feature";
+      }
       description
         "When supported, this tells the router to propagate any
-         prefixes that are attached to these community-types.
-         The value of 0 implies 'none'.";
+         prefixes that are attached to these community-types.";
     }
     leaf description {
       type string;
--- a/src/yang/ietf-bgp-types.yang
+++ b/src/yang/ietf-bgp-types.yang
@@ -359,6 +359,35 @@ module ietf-bgp-types {
       "RFC 5065, Autonomous System Configuration for BGP.";
   }
 
+  identity send-community-feature {
+    description
+      "Base identity to identify send-community feature.";
+  }
+
+  identity standard {
+    base send-community-feature;
+    description
+      "Send standard communities.";
+    reference
+      "RFC 1997: BGP Communities Attribute.";
+  }
+
+  identity extended {
+    base send-community-feature;
+    description
+      "Send extended communities.";
+    reference
+      "RFC 4360: BGP Extended Communities Attribute.";
+  }
+
+  identity large {
+    base send-community-feature;
+    description
+      "Send large communities.";
+    reference
+      "RFC 8092: BGP Large Communities Attribute.";
+  }
+
   /*
    * Typedefs.
    */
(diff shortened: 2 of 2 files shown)


mjethanandani (commented):
While the changes look good to me, I decided to modify one of the example configurations to see if we could add an example to match this code change. And that is when I see this error message:

Validating yang/example-bgp-configuration-a.1.1.xml
err : Identityref value is not derived from all its bases. (/ietf-routing:routing/control-plane-protocols/control-plane-protocol[type='ietf-bgp:bgp'][name='name:BGP']/ietf-bgp:bgp/neighbors/neighbor[remote-address='192.0.2.1']/send-community[.='standard'])

For this modified example:

[example-bgp-configuration-a.1.1.xml.zip](https://github.com/mjethanandani/ietf-bgp-yang/files/7798418/example-bgp-configuration-a.1.1.xml.zip)

Do not know if the issue is with yanglint or the code itself. We should probably validate and make sure the changes are ok.

jhaas-pfrc:
> While the changes look good to me, I decided to modify one of the example configurations to see if we could add an example to match this code change. And that is when I see this error message:
> 
> Validating yang/example-bgp-configuration-a.1.1.xml err : Identityref value is not derived from all its bases. (/ietf-routing:routing/control-plane-protocols/control-plane-protocol[type='ietf-bgp:bgp'][name='name:BGP']/ietf-bgp:bgp/neighbors/neighbor[remote-address='192.0.2.1']/send-community[.='standard'])
> 
> For this modified example:
> 
> [example-bgp-configuration-a.1.1.xml.zip](https://github.com/mjethanandani/ietf-bgp-yang/files/7798418/example-bgp-configuration-a.1.1.xml.zip)
> 
> Do not know if the issue is with yanglint or the code itself. We should probably validate and make sure the changes are ok.

I'm not sure how to execute the example, but I think you're missing the bt: qualifier.  See example here:
https://datatracker.ietf.org/doc/html/rfc7950#section-9.10.5

mjethanandani:
> > While the changes look good to me, I decided to modify one of the example configurations to see if we could add an example to match this code change. And that is when I see this error message:
> > Validating yang/example-bgp-configuration-a.1.1.xml err : Identityref value is not derived from all its bases. (/ietf-routing:routing/control-plane-protocols/control-plane-protocol[type='ietf-bgp:bgp'][name='name:BGP']/ietf-bgp:bgp/neighbors/neighbor[remote-address='192.0.2.1']/send-community[.='standard'])
> > For this modified example:
> > [example-bgp-configuration-a.1.1.xml.zip](https://github.com/mjethanandani/ietf-bgp-yang/files/7798418/example-bgp-configuration-a.1.1.xml.zip)
> > Do not know if the issue is with yanglint or the code itself. We should probably validate and make sure the changes are ok.
> 
> I'm not sure how to execute the example, but I think you're missing the bt: qualifier. See example here: https://datatracker.ietf.org/doc/html/rfc7950#section-9.10.5

To execute the example, you can run the whole build, or run just validate-and-gen-tree.sh script in the src directory, or just the command:

`% yanglint -ii -t config -p ../../yang-parameters -p ../bin -p ../bin/submodules ../../yang-parameters/ietf-network-instance@2019-01-21.yang ../bin/ietf-bgp-types@2022-01-02.yang ../bin/ietf-bgp@2022-01-02.yang yang/example-bgp-configuration-a.1.1.xml`

You are correct. It does need the prefix bt:.

mjethanandani:
I have suggested some additional changes as part of a PR. You can choose to pull in those changes into this commit, or alternatively, I can merge them directly into v13.

mjethanandani (approved):

keyur-arrcus (approved):

jhaas-pfrc:
> I have suggested some additional changes as part of a PR. Those changes update the example for testing this change. You can choose to pull in those changes into this commit, or alternatively, I can merge them directly into v13.

Let's do those changes separately.  One of the things we'll want to review is our examples and that could use some holistic work.

PR #162, opened 2022-01-03
Commits on top of large-community changes

mjethanandani:
The changes for updating the example to exercise sending of community features. @jhaas-pfrc, I have made these changes on top of your branch for large-community changes, so an approval will allow you to merge these changes into your branch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,5 @@
-FROM mjethanandani/build-yang:1.0
+FROM mjethanandani/build-yang:latest
 
-# Everything below is draft specific
-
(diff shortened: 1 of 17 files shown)


keyur-arrcus (approved):
approving.

suehares:
Do we want to include WIDE communities?  My understanding is that we would include WIDE Communities as an extension.

jhaas-pfrc:
> Do we want to include WIDE communities? My understanding is that we would include WIDE Communities as an extension.

It doesn't belong in the base model.  We've done the work so that future models that are community-like can be done as extensions.

Issue #163, opened 2022-01-04
Set consensus to true in the <rfc> tag


mjethanandani:
The current build of the draft complains:

draft-ietf-idr-bgp-model-13.xml(3): Warning: Setting consensus="true" for IETF STD document (this is not the schema default, but is the only value permitted for this type of document)

PR #164, opened 2022-01-04
Fix consensus status in the draft.

mjethanandani:
Issue #163.

I know it has more than the consensus commit, but it is more work trying to remove the fixes for the Docker build.


jhaas-pfrc (approved):
I'm not sure if our setting for consensus matters, but it doesn't hurt to make the warning go away.

Issue #168, opened 2022-01-18
Enhance policy part of the model


mjethanandani:
This from @jhaas-pfrc:

What is the list of core policy operations we want to support?

Here's the start of a list.

match on afi/safi
Prefix matching.  Do we support exact only as opposed to range?  (junos e.g. prefix-list vs. route-filter)  Do we care about wildcard mask matches?
set localpref.  Match localpref (range?)
set med. set med + aigp?  set med + igp?  Match med (range?)
match on as-path regex (whose regex?)
match on community regex (whose regex?)
match on community using set algebra (whose semantics?  intersect? wholly contained?0
as-path prepend
community manipulate (all types) - add, delete, (re)set
match/set origin
match community-count
match neighbor? (ingress only?)
set next-hop

Policy for route flap damping?

mjethanandani:
Hi @jhaas-pfrc,

> match on afi/safi 

I believe this is already supported under neighbors->neighbor->afi-safis->afi-safi->apply-policy->import-policy

mjethanandani:
Hi @jhaas-pfrc,

> set next-hop

Did you mean match next-hop or set next-hop? Either ways, I have added support for match next-hop for now. See PR#170.

jhaas-pfrc:
The list in email was constructed as the requirements rather than the todo list.
I think afi-safi is covered in the "afi-safi-in" element.

keyur-arrcus:
Can we keep the regex related commands, community manipulate and aigp commands out for now? We will need a generic regex container to be defined for getting regex related commands. Also AIGP came a bit later on so won't fit into base model if you ask me.

mjethanandani:
I have done an analysis of the list that @jhaas-pfrc, put together as part of this issue. With the combination of the two PRs that are currently under review, @keyur-arrcus, question about what he thinks should be addressed, I believe the only thing we are missing is 'match neighbor'. I will initiate a PR for that.

If we agree, then I can close this issue once the two PRs have been approved.

jhaas-pfrc:
I think it's irresponsible to leave out community and as-path matching as they're two fundamental bits of things we need.

I agree deciding what to do about regex will take a bit of thinking.
Community matching may not only use regex but also community boolean algebra that's common in most implementations.

mjethanandani:
> I think it's irresponsible to leave out community and as-path matching as they're two fundamental bits of things we need.

But we do support match-community-set and match-as-path-set already. Am I missing something?

PR #169, opened 2022-01-18
Add support for match prefix-set

mjethanandani:
This is one in a series of commits to address the issues listed in issue #168.


src/yang/ietf-bgp-policy.yang
```diff
@@ -21,7 +21,7 @@ module ietf-bgp-policy {
   organization
     "IETF IDR Working Group";
   contact
-    "WG Web:   <http://tools.ietf.org/wg/idr>
+    "WG Web:   <http://datatracker.ietf.org/wg/idr>
```
jhaas-pfrc, line 24: Repeating comments from before, it's ideal to keep unrelated bits of diff from pull requests.  This probably should go under a separate issue to update the copyright dates.

src/yang/ietf-bgp-policy.yang
```diff
@@ -61,31 +61,33 @@ module ietf-bgp-policy {
...
+   * typedef statements
+   */
 
   typedef bgp-set-community-option-type {
     type enumeration {
       enum add {
         description
           "Add the specified communities to the existing
-           community attribute";
+           community attribute.";
```
jhaas-pfrc, line 73: Same comment for the unrelated punctuation cleanup

src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +253,92 @@ module ietf-bgp-policy {
...
+
+          leaf name {
+            type string;
+            description
+              "Name of the prefix set used to reference the set in
+               match conditions.";
+          }
+
+          leaf mode {
+            type enumeration {
```
jhaas-pfrc, line 275: What does our extension model look like for updating this?
For example, while matching ipv4 and ipv6 prefixes is a core common application, what happens if you want your policy to match a vpnv4 prefix?  A rt-constrain prefix?  Etc.

src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +253,92 @@ module ietf-bgp-policy {
...
+                  "Prefix set contains mixed IPv4 and IPv6 prefixes";
+              }
+            }
+            description
+              "Indicates the mode of the prefix set, in terms of
+               which address families (IPv4, IPv6, or both) are
+               present. The mode provides a hint, but the device
+               must validate that all prefixes are of the indicated
+               type, and is expected to reject the configuration if
+               there is a discrepancy.  The 'mexed' mode may not be
```
jhaas-pfrc, line 295: typo - mexed

src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +253,92 @@ module ietf-bgp-policy {
...
+                 prefix may be either IPv4 or IPv6, most
+                 implementations require all members of the prefix
+                 set to be the same address family. Mixing address
+                 types in the same prefix set is likely to cause an
+                 error.";
+            }
+
+            leaf masklength-range {
+              type string {
+                pattern '^(([0-9]+\.\.[0-9]+)|exact)$';
```
jhaas-pfrc, line 325: What do people feel about "longer" and "or-longer" as additional easy forms?

jhaas-pfrc: Also, regexes can't count.  General advice is that the first part of the range must be less than or equal to the second part of the range.  How do we want to note that?

jhaas-pfrc (changes requested):

src/yang/ietf-bgp-policy.yang
keyur-arrcus: We will just have you augment it. :(

src/yang/ietf-bgp-policy.yang
keyur-arrcus: That is correct. my suspicion is that the range check would be hard. *If* we want, we can move it in the validation part.

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: Right.  See prior discussions about the issues with augmenting for enumerations.  You can end up with inconsistent enumerations as a side effect.

Mahesh had a pattern that moved us to identities for the community types.  That might be appropriate here?

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: To make this a concrete request:
add named options longer and or-longer
add to the description that number first number must be <= second number and that first number must be >= the prefix-length.

mjethanandani:
@jhaas-pfrc, do the latest set of changes address your comments?

src/yang/ietf-bgp-policy.yang
mjethanandani: The tools are flagging the Copyright date as an issue, and they use the last modified date to decide whether the date should be updated.

keyur-arrcus (approved):
looks good so far.

src/yang/ietf-bgp-policy.yang
mjethanandani: > What do people feel about "longer" and "or-longer" as additional easy forms?

I added "longer" and "or-longer" as additional options in the mask length options, but examples on the web seem to indicate use of "le" or "ge". What is preferable?

mjethanandani:
BTW, these changes have been validated with the example in PR #175.

src/yang/ietf-bgp-policy.yang
```diff
@@ -272,6 +300,95 @@ module ietf-bgp-policy {
...
+          key "ip-prefix masklength-range";
+          description
+            "List of prefixes in the prefix set.";
+
+          leaf ip-prefix {
+            type inet:ip-prefix;
+            mandatory true;
+            description
+              "The prefix member in CIDR notation. While the
+                prefix may be either IPv4 or IPv6, most
```
jhaas-pfrc, line 352: Indentation issue.

src/yang/ietf-bgp-policy.yang
```diff
@@ -272,6 +300,95 @@ module ietf-bgp-policy {
...
+        }
+      }
+
+      container prefixes {
+        description
+          "Enclosing container for the list of prefixes in a policy
+           prefix list.";
+
+        list prefix {
+          key "ip-prefix masklength-range";
```
jhaas-pfrc, line 343: A question for Keyur is do we think that implementations that support disjoint overlapping prefixes with range have consistent implementation for lookup.

jhaas-pfrc (approved):

PR #170, opened 2022-01-20
Add support for match next-hop

mjethanandani:
Although issue #168 calls for set next-hop, this commit is for match next-hop.

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -259,16 +259,13 @@ module ietf-bgp-policy {
             "List of defined next-hop sets for use in policies.";
 
           leaf name {
-            type string {
-              length "1..79";
-            }
+            type string;
             mandatory true;
             description
               "Name of the next-hop set.";
           }
           leaf-list next-hop {
             type bgp-next-hop-type;
-            max-elements 32;
             description
               "List of IP addresses in the next-hop set.";
           }
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -266,7 +266,7 @@ module ietf-bgp-policy {
             description
               "Name of the next-hop set.";
           }
-          leaf-list address {
+          leaf-list next-hop {
             type bgp-next-hop-type;
             max-elements 32;
             description
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -248,6 +248,32 @@ module ietf-bgp-policy {
           }
         }
       }
+      container next-hop-sets {
+        description
+          "Definition of a list of IPv4 or IPv6 next-hops which can
+           be matched in a routing policy.";
+
+        list next-hop-set {
+          key "name";
+          description
+            "List of defined next-hop sets for use in policies.";
+
+          leaf name {
+            type string {
+              length "1..79";
+            }
+            mandatory true;
+            description
+              "Name of the next-hop set.";
+          }
+          leaf-list address {
+            type bgp-next-hop-type;
+            max-elements 32;
+            description
+              "List of IP addresses in the next-hop set.";
+          }
+        }
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +248,32 @@ module ietf-bgp-policy {
...
+           be matched in a routing policy.";
+
+        list next-hop-set {
+          key "name";
+          description
+            "List of defined next-hop sets for use in policies.";
+
+          leaf name {
+            type string {
+              length "1..79";
```
jhaas-pfrc, line 263: Why is this being constrained?

src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +248,32 @@ module ietf-bgp-policy {
...
+            }
+            mandatory true;
+            description
+              "Name of the next-hop set.";
+          }
+          leaf-list address {
```
jhaas-pfrc, line 269: Should this simply be "next-hop"?
I don't know that we have a pattern for the policy sets yet.  For example, if the match container is «thing»-set, then the individual elements perhaps should be «thing»?

src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +248,32 @@ module ietf-bgp-policy {
...
+            type bgp-next-hop-type;
+            max-elements 32;
```
jhaas-pfrc, line 271: Why is this constrained?

src/yang/ietf-bgp-policy.yang
```diff
@@ -248,6 +248,32 @@ module ietf-bgp-policy {
...
+          leaf name {
+            type string {
+              length "1..79";
+            }
+            mandatory true;
+            description
+              "Name of the next-hop set.";
+          }
+          leaf-list address {
+            type bgp-next-hop-type;
```
jhaas-pfrc, line 270: I think this type will _probably_ be fine.  It's scoped as the "no-zone" one.
The thing I'm concerned about is implementations that only want matching address family elements for a given session type.  For example, for ipv4 unicast we should only have nexthops that are ipv4.

Juniper's implementation, for example, had a period where matched nexthops were a requirement.  Later, when support for the various v4 over v6 or vice versa features were introduced, a mix was possible.

This is a challenging IETF implementation-neutral question.  It may simply be easiest to not constrain for address family matching and let implementations enforce it at commit time.

I'd appreciate Keyur's thinking here too.

src/yang/ietf-bgp-policy.yang
mjethanandani: Good question. OC has this in their model.

src/yang/ietf-bgp-policy.yang
mjethanandani: What we have for a pattern is

```
container blah-sets {

  list blah-set {

    leaf-list <whatever best describes the list> {

    }
  }
}
```

which we could change to:

```
container blah-sets {

  list blah-set {

    leaf-list blah {

    }
  }
}
```

But we would have to do it for all cases in that case.

src/yang/ietf-bgp-policy.yang
mjethanandani: Again, this is what is in the OC model.

src/yang/ietf-bgp-policy.yang
keyur-arrcus: IMHO, we should perform a check of same session supporting same nexthop-type at the back end as part of validation. Most implementations today support v4 over v6 and vice versa.

keyur-arrcus: One more comment.  since its match next-hop I would suggest keeping next-hop type as inept:ip-address as opposed to no-zone-ip-address?

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: We aren't OC.  IETF shouldn't be setting arbitrary limits on every BGP user on the planet.  Vendors are welcome to add their own constraints here.

I'm still angry at all of the MPLS MIBs that have Cisco limitations in them that mucked up implementations I've worked on.

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: > One more comment. since its match next-hop I would suggest keeping next-hop type as inept:ip-address as opposed to no-zone-ip-address?

Zone is not present in the PDUs; it's an inferred property of the session.  That's why it should be excluded.

PR #175, opened 2022-01-25
Add a policy example

mjethanandani:
This is my attempt to add a example on how to configure a policy using the BGP policy model.

The specific example that I am trying to follow is documented at:

https://community.cisco.com/t5/networking-documents/bgp-match-community-quot-statement-used-in-route-map/ta-p/3149898

I have several questions on this example. 

- How is the policy attached? Is it through the apply-policy->import-policy or apply-policy->export-policy?
- Is a (bgp)-defined-set the same as a route-map?
- How is prefix-mode decided?
- What should the masklength-range be in this case?
- route-map includes a 'set community' command? But 'set-community' under 'actions' has a 'community-set-ref' which points back to 'community-set'. Does it mean that I should be defining 11:11 as a community-set?
- What would be a good 'name' under policy-definition?
- What would be a good 'name under 'statement'?

### DIFF
--- a/draft/draft-ietf-idr-bgp-model.xml
+++ b/draft/draft-ietf-idr-bgp-model.xml
@@ -945,14 +945,11 @@ INSERT_TEXT_FROM_FILE(../src/yang/example-bgp-configuration-a.1.3.xml,68)
         R1, R2, and R3, configured with AS value of 1, 2 and 3 
         respectively. R1 advertises 5 prefixes to R2 and R3, as
         shown below.</t>
-          <dl newline="true" spacing="normal" indent="3">
-	    <dt>1.1.1.1/32 and 2.2.2.2/32</dt>
-	    <dd>community 11:11</dd>
-	    <dt>3.3.3.3/32 and 4.4.4.4/32</dt>
-	    <dd>community 11:11 22:22</dd>
-	    <dt>5.5.5.5/32</dt>
-	    <dd>community 33.33</dd>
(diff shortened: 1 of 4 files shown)


mjethanandani:
Attaching the text version of the draft as it is easier to read. [draft-ietf-idr-bgp-model-13.txt](https://github.com/mjethanandani/ietf-bgp-yang/files/7973703/draft-ietf-idr-bgp-model-13.txt)

jhaas-pfrc (commented):
It looks like your commit here pulls in stuff errantly from the prefix-set work.

mjethanandani:
> It looks like your commit here pulls in stuff errantly from the prefix-set work.

I am looking at the following configuration statements for why prefix-set is included. Do you think there is another way to define these sets?

```
ip prefix-list L0&L1 seq 5 permit 1.1.1.1/32

ip prefix-list L0&L1 seq 10 permit 2.2.2.2/32

!

ip prefix-list L2&L3 seq 5 permit 3.3.3.3/32

ip prefix-list L2&L3 seq 10 permit 4.4.4.4/32

!

ip prefix-list L4 seq 5 permit 5.5.5.5/32
```

mjethanandani:
A question for everyone. Should the set-community action have a default value (of add)? Currently, the user can specify a prefix-set, but not specify any 'option' of add, remove or replace on it.

jhaas-pfrc (approved):

PR #187, opened 2022-02-28
Add support for med-plus-igp

mjethanandani:
When reviewing please check to see whether an implementation would have access to med and igp cost values. I see med in rib entries, but is aigp-metric the igp cost value?


suehares:
Approve with change #187

mjethanandani:
Reopening to merge the changes.

PR #189, opened 2022-03-03
Fix prefix set definition and the corresponding match

mjethanandani:
As currently defined, prefix-sets are not a child of bgp-defined-sets, and prefixes are not a child of list prefix-set. 

The former I believe should be a child, unless we believe that the prefix-sets can be used outside of bgp(-defined-sets). The latter I believe leads to a prefix-set definition that contains no prefixes.

But since these code was reviewed before, and these changes are coming at the last minute, I want us to be careful reviewing this change before accepting it.


suehares (approved):
works for me.

jhaas-pfrc (approved):

PR #197, opened 2022-04-04
Add support for med-plus-igp

mjethanandani:


keyur-arrcus (approved):
Looks good to me.

jhaas-pfrc:
> Looks good to me.

@keyur-arrcus does Cisco or Arrcus do med-plus-igp as policy?  I wasn't able to find a Cisco command that covers this.

Juniper and Huawei do this as path-selection options, not policy.

mjethanandani:
> > Looks good to me.
> 
> @keyur-arrcus does Cisco or Arrcus do med-plus-igp as policy? I wasn't able to find a Cisco command that covers this.
> 
> Juniper and Huawei do this as path-selection options, not policy.

Hi @jhaas-pfrc , you are right. This is implemented as a route selection option, and is already implemented both at a global level and at a per aft/safi level. Will close this out.

PR #198, opened 2022-04-04
Fix the regex pattern to comply with RFC 7950

mjethanandani:
The current pattern for the regex is not RFC 7950 compliant, as it follows the POSIX standard, instead of the W3C standard which is what RFC 7950 follows.


keyur-arrcus (approved):
Looks ok.

jhaas-pfrc (approved):
I wasn't aware that regex are implicitly anchored. This will have to go into my review criteria.

Thanks.

Issue #214, opened 2022-05-20
Update TCP-AO example


mjethanandani:
The current example in the Appendix assumes that TCP YANG module holds the definition of parameters for TCP-AO. The TCP YANG model has updated the authentication parameters to be moved under an augmentation of the Key Chain model. Therefore the example needs to be updated to reflect that change.

PR #215, opened 2022-05-20
Update AO example

mjethanandani:
See #214 for details.

### DIFF
--- a/src/yang/example-bgp-configuration-a.1.2.xml
+++ b/src/yang/example-bgp-configuration-a.1.2.xml
@@ -3,8 +3,58 @@
     xmlns="urn:ietf:params:xml:ns:yang:ietf-key-chain">
   <key-chain>
     <name>bgp-key-chain</name>
+    <description>"An example of TCP-AO configuration for BGP"</description>
+    <key>
+      <key-id>55</key-id>
+      <lifetime>
+        <send-lifetime>
+          <start-date-time>2017-01-01T00:00:00Z</start-date-time>
+          <end-date-time>2017-02-01T00:00:00Z</end-date-time>
+        </send-lifetime>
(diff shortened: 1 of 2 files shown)


src/yang/example-bgp-configuration-a.1.2.xml
```diff
@@ -3,8 +3,58 @@
...
+          <end-date-time>2017-02-01T00:00:00Z</end-date-time>
+        </send-lifetime>
+        <accept-lifetime>
+          <start-date-time>2016-12-31T23:59:55Z</start-date-time>
+          <end-date-time>2017-02-01T00:00:05Z</end-date-time>
+        </accept-lifetime>
+      </lifetime>
+      <crypto-algorithm>hmac-sha-1</crypto-algorithm>
+      <key-string>
+	<keystring>testvector</keystring>
```
jhaas-pfrc, line 21: Indentation is off here.

jhaas-pfrc (changes requested):
Multiple places indentation is off. (Viewed in Firefox)

src/yang/example-bgp-configuration-a.1.2.xml
mjethanandani: Hmm. I am not seeing any indentation issues when viewed within Emacs and here in Safari.

jhaas-pfrc: The issue in the file is that you have some lines with tabs in them rather than whitespace.  This will render wrong and catch our eye in the file review.

Since these are intended to be embedded as "artwork", removing the tabs is important.

mjethanandani: Got it.

PR #218, opened 2022-06-17
How to maintain extended-community datatype

mjethanandani:
This PR addresses the issue of how to make the ext-community-type extensible. It does it by
- adding a string definition called 'raw', that allows for new and experimental sub-types to be defined, and the 'raw' string is followed by 8 bytes of hex values.
- moving the module to a IANA maintained module.

### DIFF
--- a/src/yang/iana-bgp-types.yang
+++ b/src/yang/iana-bgp-types.yang
@@ -514,7 +514,10 @@ module iana-bgp-types {
       }
     }
     description
-      "Type definition for extended community attributes";
+      "Type definition for extended community attributes.
+       It includes a way to specify a 'raw' string that
+       is followed by 8 bytes of octet string to support
+       new and experimental type definitions.";
     reference
       "RFC 4360 - BGP Extended Communities Attribute";
   }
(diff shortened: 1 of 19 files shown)


src/yang/ietf-bgp-rib-attributes.yang
```diff
@@ -223,6 +223,15 @@ submodule ietf-bgp-rib-attributes {
       reference
         "RFC 4360 - BGP Extended Communities Attribute";
     }
+
+    leaf-list ext-community-raw {
+      type string {
+        // raw with 8 octetes
```
jhaas-pfrc, line 229: octets

jhaas-pfrc (changes requested):
bgp-ext-community-type also needs an option for the raw format as well.

src/yang/iana-bgp-types.yang
```diff
@@ -497,9 +507,17 @@ module ietf-bgp-types {
               + '(6[0-5][0-5][0-3][0-5]|[1-5][0-9]{4}|'
               + '[1-9][0-9]{1,4}|[0-9])';
       }
+
+      type string {
+        // raw with 8 octetes
```
jhaas-pfrc, line 512: octets

jhaas-pfrc:
> 

My comment about this needing to be in the bgp-ext-community-type was in error. It's in the diff.

PR #266, opened 2023-01-04
Dump capabilities

### LINKED ISSUE #248: Dump capabilities better
In ietf-bgp.yang we have a leaf-list of "negotiated-capabilities" for our neighbor.  This is a handy thing to have, but we typically also want to have the list of capabilities we have locally configured and will have sent, and the set that is received from the remote bgp peer.

We don't want to make this configuration state, it's only debugging operational state.

jhaas-pfrc:
The OC-inherited model only showed "negotiated capabilities".  There's an operational need to be able to show the sent/received capabilities and their  capability-specific data.

Closes #248


jhaas-pfrc:
Branch based on other pending pull request.  Closing until that one is addressed.

PR #272, opened 2023-01-05
Issue #254 Remove extended bit from path unknown path attribute

### LINKED ISSUE #254: grouping bgp-unknown-attr-top has leaf extended and possibly shouldn't
The grouping bgp-unknown-attr-top represents the BGP unknown attributes that aren't otherwise present in the model.

The leaf "extended" is meant to cover the path attribute flag "extended" that is used to decide whether parsing needs one or two bytes of data.

There is a leaf that carries the length. 

There's a narrow operational case where it might be useful debugging to note that a value < 256 bytes was carried in two bytes of attr-length, but that's really the only reason to keep this around.  It otherwise just creates noise.

jhaas-pfrc:
The unknown-attr model IETF inherited from the OpenConfig model included a boolean flag for the "extended" bit in the Path Attribute flags.  This flag is used to indicate that the length field after the Path Code is either one or two octets in length.

This flag is not required to be preserved by the protocol since the length itself is normally enough to indicate whether two-octet encoding is required.

There is an edge case that length < 256 octets for a path attribute may be encoded as a two-octet field.  While this isn't the usual desired encoding, it's still valid and introduces no headaches.  RFC 4271 and RFC 7606 are silent about such a case.

Since implementations aren't required to preserve the received Extended flag, remove it from this model and document the reason in the length field.

Closes #254

### DIFF
--- a/src/yang/ietf-bgp-rib-attributes.yang
+++ b/src/yang/ietf-bgp-rib-attributes.yang
@@ -128,17 +128,6 @@ submodule ietf-bgp-rib-attributes {
             "RFC 4271 - A Border Gateway Protocol 4 (BGP-4)";
         }
 
-        leaf extended {
-          type boolean;
-          description
-            "Defines whether the attribute length is one octet
-             (if set to false) or two octets (if set to true). Set in
-             the fourth high-order bit of the BGP attribute flags
-             octet.";
-          reference
-            "RFC 4271 - A Border Gateway Protocol 4 (BGP-4)";
-        }
-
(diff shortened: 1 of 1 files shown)


jhaas-pfrc (commented):
I would like explicit ack on this from Keyur and Sue.

mjethanandani (approved):

jhaas-pfrc:
Keyur says that Arrcus caches the extended bit, so there's at least one user.  This PR will be dropped.

PR #273, opened 2023-01-05
Issue #255 Address structure of secure-session

### LINKED ISSUE #255: Structure of secure-session and secure-session-enable
We have the following structure in ietf-bgp.yang:

```
      leaf secure-session-enable {
            type boolean;
            default "false";
            description
              "Does this session need to be secured?";
          }

          container secure-session {
            /* XXX JMH - why is this constraint mandatory? also, why not have the enable in this container? */
            when "../secure-session-enable = 'true'";
            description
              "Container for describing how a particular BGP session
               is to be secured.";
```

As noted by my comment in secure-session, why is that constraint mandatory?
Also, why is the enable outside of the container?

I suspect we want to move secure-session-enable into the secure-session container as "enable".  This also permits configurable of security parameters without necessarily having them enabled at the moment, although some might consider that a bug rather than a feature.  (You configured it... did you even bother to turn it on?

jhaas-pfrc:
Issue #255 Address structure of secure-session
    
The enabled boolean didn't reside in the container with the secure-session
parameters themselves.  The constraint on the existing options had a constraint
that you couldn't configure options without setting the enabled status to true.
This is the opposite of how you'd normally want this to operate.
    
Move the enabled into the secure-session container.
Create a must constraint that when enabled, options MUST be configured.
To support the must constraint, put the current choice option into new container "options" to provide a XPath testable element.
Move secure-session into the transport container and remove redundant copy under peer-groups.
Fix the examples using secure-session.

Closes #255

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -401,6 +401,20 @@ submodule ietf-bgp-common {
               "RFC 5925: The TCP Authentication Option.";
           }
 
+          case ipsec {
+            leaf sa {
+              type string;
+              description
+                "Security Association (SA) name.";
+            }
+            description
+              "Currently, the IPsec/IKE YANG model has no
+               grouping defined that this model can use. When
+               such a grouping is defined, this model can import
+               the grouping to add the key parameters
+               needed to kick of IKE.";
+          }
+
           description
             "Choice of authentication options.";
         }
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -9,6 +9,11 @@ submodule ietf-bgp-common {
     reference
       "RFC XXXX: BGP Model for Service Provider Network.";
   }
+  import ietf-key-chain {
+    prefix key-chain;
+    reference
+      "RFC 8177: YANG Data Model for Key Chains";
+  }
   import ietf-inet-types {
     prefix inet;
     reference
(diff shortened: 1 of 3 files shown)


jhaas-pfrc:
> Noticed that you dropped IPSec as an authentication option. Is that on purpose?

That's an error in the change.  That's why we have reviews. :-)

jhaas-pfrc:
> Noticed that you dropped IPSec as an authentication option. Is that on purpose?

Looking at the diff, the issue was that ipsec was not consistently present in both places we had secure-session.  So, an accidental bug fix is happening. :-)

jhaas-pfrc:
@mjethanandani I believe I've addressed the open issues here. Could you verify?

src/yang/ietf-bgp-common.yang
```diff
@@ -340,6 +345,82 @@ submodule ietf-bgp-common {
...
+          "Options for securing the BGP session.";
+        choice option {
+          case ao {
+            leaf ao-keychain {
+              type key-chain:key-chain-ref;
+              description
+                "Reference to the key chain that will be used by this
+                 model. Applicable for TCP-AO and TCP-MD5 only";
+              reference
+                "RFC 8177: YANG Key Chain.";
```
mjethanandani, line 377: The title of 8177 is different from the title above in the import statement.

src/yang/ietf-bgp-common.yang
```diff
@@ -340,6 +345,82 @@ submodule ietf-bgp-common {
...
+          }
+
+          case md5 {
+            leaf md5-keychain {
+              type key-chain:key-chain-ref;
+              description
+                "Reference to the key chain that will be used by this
+                 model. Applicable for TCP-AO and TCP-MD5 only";
+              reference
+                "RFC 8177: YANG Key Chain.";
```
mjethanandani, line 394: Same here.

src/yang/ietf-bgp-common.yang
```diff
@@ -340,6 +345,82 @@ submodule ietf-bgp-common {
...
+            }
+            description
+              "Uses TCP-AO to secure the session. Parameters for
+               those are defined as a grouping in the TCP YANG
+               model.";
+            reference
+              "RFC 5925 - The TCP Authentication Option.";
```
mjethanandani, line 384: While you are here, can you fix the format of this reference statement.

mjethanandani:
Sorry for not catching some of this in the initial review.

src/yang/ietf-bgp-common.yang
jhaas-pfrc: Fixed in this file.

src/yang/ietf-bgp-common.yang
jhaas-pfrc: Fixed in this file.

src/yang/ietf-bgp-common.yang
jhaas-pfrc: Fixed in this file.

src/yang/ietf-bgp-common.yang
```diff
@@ -374,14 +374,14 @@ submodule ietf-bgp-common {
...
-                "RFC 8177: YANG Key Chain.";
+                "RFC 8177: YANG Data Model for Key Chains.";
             }
             description
               "Uses TCP-AO to secure the session. Parameters for
                those are defined as a grouping in the TCP YANG
                model.";
             reference
-              "RFC 5925 - The TCP Authentication Option.";
+              "RFC 5925:he TCP Authentication Option.";
```
mjethanandani, line 384: s/he/ The/

mjethanandani (approved):

PR #274, opened 2023-01-05
Issue #257 - move ttl-security under transport container

### LINKED ISSUE #257: leaf ttl-security should likely move to containers transport
leaf ttl-security is floating by itself in the neighbor container.  It should likely be rolled into the transport containers since it covers transport behavior.

Perhaps also consider moving the secure-session container next to transport?

jhaas-pfrc:
Closes #257

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -340,6 +340,17 @@ submodule ietf-bgp-common {
          of an interface.";
     }
 
+    leaf ttl-security {
+      if-feature "bt:ttl-security";
+      type uint8;
+      default "255";
+      description
+        "BGP Time To Live (TTL) security check.";
+      reference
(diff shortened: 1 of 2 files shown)


mjethanandani (approved):
The changes for ttl-security look good to me.

However, you have a comment in the issue that we should consider moving secure-session container under transport. Today, just like ttl-security, it exists both at a neighbor level and at a peer-group level. Also, at the peer-group level, there is IPSec option, which does not exist at a neighbor level.

jhaas-pfrc:
> The changes for ttl-security look good to me.
> 
> However, you have a comment in the issue that we should consider moving secure-session container under transport. Today, just like ttl-security, it exists both at a neighbor level and at a peer-group level. Also, at the peer-group level, there is IPSec option, which does not exist at a neighbor level.

secure-session is moved under transport in a different PR.
The ipsec stuff was cleaned up in another PR.

Issue #263, opened 2023-01-04
Normalize reference format


jhaas-pfrc:
The desired reference format in the draft is:
"RFC XXXX: <Title of the RFC> [Section number].";

Lots of places in the document aren't in this format.

suehares:
Format works for me.

PR #279, opened 2023-01-05
Issue #277 - Hold and established time changes

### LINKED ISSUE #277: Move negotiated-hold-time and fsm-established-time to timers container

jhaas-pfrc:
Move the negotiated-hold-time leaf to the timers container.

Remove the fsm-established-time leaf.  The leaf, last-established, covers the event when we transition in and out of Established.  The timer that's being deleted can be derived from the last-established timestamp and the current session-state.  This is also more consistent with the OpenConfig model.

Closes #277

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -121,6 +121,15 @@ submodule ietf-bgp-common {
         "RFC 4271, Section 4.2.
          RFC 4271, Section 10.";
     }
+    leaf negotiated-hold-time {
+      type uint16;
+      config false;
+      description
+        "The negotiated hold-time for the BGP session";
+      reference
+        "RFC 4271, Section 4.2.
+         RFC 4271, Section 10.";
+    }
     leaf keepalive {
       type uint16 {
         range "0..21845";
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -590,6 +590,8 @@ module ietf-bgp {
                the difference between this value and the current time
                in UTC (assuming the session is in the Established
                state, per the session-state leaf).";
+            reference
+              "RFC 4271, Section 8.";
           }
           leaf-list negotiated-capabilities {
             type identityref {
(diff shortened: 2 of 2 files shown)


src/yang/ietf-bgp-common.yang
```diff
@@ -121,6 +121,15 @@ submodule ietf-bgp-common {
         "RFC 4271, Section 4.2.
          RFC 4271, Section 10.";
     }
+    leaf negotiated-hold-time {
+      type uint16;
+      config false;
+      description
+        "The negotiated hold-time for the BGP session";
+      reference
+        "RFC 4271, Section 4.2.
```
mjethanandani, line 130: If we agree on the pattern for the reference statement, described in issue #263, can these statements be updated to reflect the pattern?

src/yang/ietf-bgp.yang
```diff
@@ -590,6 +590,8 @@ module ietf-bgp {
                the difference between this value and the current time
                in UTC (assuming the session is in the Established
                state, per the session-state leaf).";
+            reference
+              "RFC 4271, Section 8.";
```
mjethanandani, line 594: Same here.

src/yang/ietf-bgp.yang
```diff
@@ -620,21 +616,6 @@ module ietf-bgp {
...
-              "This timer indicates how long (in
-               seconds) this peer has been in the
-               established state or how long
-               since this peer was last in the
-               established state.  It is set to zero when
-               a new peer is configured or when the router is
-               booted.";
-            reference
-              "RFC 4271, Section 8.";
-          }
```
mjethanandani, line 637: Good catch!

src/yang/ietf-bgp-common.yang
jhaas-pfrc: Did this and a few more since I was in there.

src/yang/ietf-bgp.yang
jhaas-pfrc: same

mjethanandani (approved):

PR #282, opened 2023-01-05
Issue #253 - Remove redundant bfd container, change local-address

### LINKED ISSUE #251: interfaces/bfd - why is it there?
What implementations of BGP with BFD have a interfaces/bfd configuration?  While this makes sense from the BFD module standpoint, I don't think it belongs in BGP.

This lives currently in ietf-bgp.yang.

### LINKED ISSUE #253: transport/local-address and transport-bfd seem redundant
In ietf-bgp-common.yang, container transport has leaf local-address and container bfd.
The hierarchy has these same nodes present outside of the transport container.

Are these redundant?  If so, they should be removed.

jhaas-pfrc:
There was a redundant bfd container left after cleanup. It has been removed. 

bgp/[...]/transport/local-address appears redundant, but in the current model it is used to configure the local-address.  The
bgp/neighbors/transport/local-address is currently operational-only state.

Removed the bgp/interfaces container.

Moved the local-address node up to the top of transport.

Closes #253
Closes #251

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -202,7 +202,6 @@ submodule ietf-bgp-common {
          this differs from the global BGP router autonomous system
          number.";
     }
-
     leaf remove-private-as {
       type bt:remove-private-as-option;
       description
@@ -308,6 +307,20 @@ submodule ietf-bgp-common {
       "Configuration parameters relating to the transport protocol
        used by the BGP session to the peer.";
 
+    leaf local-address {
+      type union {
+        type inet:ip-address;
+        type leafref {
+          path "../../../../interfaces/interface/name";
+        }
+      }
+      description
+        "Set the local IP (either IPv4 or IPv6) address to use for
+         the session when sending BGP update messages. This may be
+         expressed as either an IP address or reference to the name
+         of an interface.";
+    }
+
     leaf tcp-mss {
       type tcp:mss;
       description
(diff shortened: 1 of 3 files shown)


src/yang/ietf-bgp-common.yang
```diff
@@ -308,6 +307,20 @@ submodule ietf-bgp-common {
       "Configuration parameters relating to the transport protocol
        used by the BGP session to the peer.";
 
+    leaf local-address {
+      type union {
+        type inet:ip-address;
+        type leafref {
+          path "../../../../interfaces/interface/name";
```
mjethanandani, line 314: Rather than using a relative path, let us use an absolute path by defining this as

'type if:interface-ref'

and importing the ietf-interfaces module in ietf-bgp-common.

mjethanandani (changes requested):

src/yang/ietf-bgp-common.yang
jhaas-pfrc: Changed.

src/yang/ietf-bgp-common.yang
```diff
@@ -9,6 +9,9 @@ submodule ietf-bgp-common {
     reference
       "RFC XXXX: BGP Model for Service Provider Network.";
   }
+  import ietf-interfaces {
+    prefix if;
+  }
```
mjethanandani, line 14: Please add a reference statement for this import statement.

mjethanandani (approved):

PR #283, opened 2023-01-05
Issue #245 - policy inequality match for local-pref and med

### LINKED ISSUE #245: localpref and med policy requires inequality operations.
Currently we have med-eq and local-pref-eq.  These things operate on 32-bit spaces and should permit matches on less-than and greater-than operations as well.

See the examples in community-count that may be appropriate.

jhaas-pfrc:
Replace exact matches for local-pref-eq and med-eq with full <=, >=, = operations.

Closes #245

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -292,11 +292,92 @@ module ietf-bgp-policy {
       description
         "Top-level container for BGP specific policy conditions.";
 
-      leaf med-eq {
-        type uint32;
+      container local-pref {
         description
-          "Condition to check if the received MED value is equal to
-           the specified value.";
+          "Value and comparison operations for conditions based on
+           the value of the BGP LOCAL_PREF.";
+        reference
+          "RFC 4271: Section 5.1.5.";
+
+        leaf value {
+          type uint32;
+          description
+            "BGP LOCAL_PREF value for comparison to the entry in the
+             BGP route.";
+        }
+
+        choice operation {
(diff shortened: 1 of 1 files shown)


mjethanandani (commented):
Did you still want to review this if you are moving towards moving the operations into a grouping?

jhaas-pfrc:
> Did you still want to review this if you are moving towards moving the operations into a grouping?

It's staged as two separate issues and reviews to permit separate discussions.  (Although that'd require someone besides you and I to be paying any real attention.)

Minimally, we already have the pattern for inequalities that is copy and paste that could use refactoring.  That's one piece of work.

The second piece of work is whether that inequality work is sufficient for local-pref and med.  I think I may be leaning toward needing something that supports ranges, as noted in the other PR.  I'll be doing a survey of a few policy mechanisms today to try to see how that's accomplished in various implementations.

jhaas-pfrc:
Reviewing a few of the user manuals for implementations, matching on med and local-pref is inconsistently supported.  Cisco RPL supports the inequalities.  Juniper aspirationally wants to, but we'll deviate for the time being.

No one that I can see supports the necessary boolean algebra behavior trivially that would permit range matches. So, that part of the work is likely out of scope.

mjethanandani (approved):

jhaas-pfrc:
Reviewed this with the other IDR chairs.  Given that there is a need to have strong consistency in policy, we don't want individual augmentations to add in comparators if we can help it.  So, we'll proceed with the inequalities.

Issue #26, opened 2019-06-18
Support for route flap damping


mjethanandani:
The model currently allows for enabling/disabling of Route Flap Damping (RFD). However, there are many parameters that need to be provided for damping to work correctly. Those parameters seem to be vendor specific.

One suggestion is for this model to allow for enable/disable of damping, but leave it to vendor specific augmentations to provide the attributes that need to set for damping to work correctly.

mjethanandani:
Need to bring this up with WG to decide what form of damping the WG would like to see supported. GateD, Cisco IOS-XR, CIsco-XE, Juniper or Ericsson.

mjethanandani:
Added support for damping in -07.

PR #285, opened 2023-01-05
Issue #284 - Refactor equality operations from policy

### LINKED ISSUE #284: Refactor eq/lt-eq/gt-eq comparator in policy module
The choice code for doing <=,=,>= comparisons is highly redundant.  It's easy to refactor it via a grouping.

jhaas-pfrc:
Closes #284

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -149,7 +149,8 @@ module ietf-bgp-policy {
   }
 
   grouping equality-operator {
-    description "";
+    description
+      "Grouping used for common equality operations in BGP policy.";
     choice operation {
       case eq {
         leaf eq {
@@ -329,7 +330,8 @@ module ietf-bgp-policy {
           "Value and comparison operations for conditions based on
            the value of the BGP LOCAL_PREF.";
         reference
-          "RFC 4271: Section 5.1.5.";
+          "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section
+           5.1.5.";
 
         leaf value {
           type uint32;
@@ -347,7 +349,8 @@ module ietf-bgp-policy {
           "Value and comparison operations for conditions based on
            the value of the BGP MULTI_EXIT_DISC (MED)";
         reference
-          "RFC 4271: Section 5.1.4.";
+          "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section
+           5.1.4.";
 
         leaf value {
           type uint32;
@@ -432,7 +435,7 @@ module ietf-bgp-policy {
            The as-path-length SHALL be calculated and SHALL follow
            RFC 4271 rules.";
         reference
-          "RFC 4271: BGP-4.";
+          "RFC 4271: A Border Gateway Protocol 4 (BGP-4).";
 
         leaf as-path-length {
           type uint32;
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -292,11 +292,92 @@ module ietf-bgp-policy {
       description
         "Top-level container for BGP specific policy conditions.";
 
-      leaf med-eq {
-        type uint32;
+      container local-pref {
         description
-          "Condition to check if the received MED value is equal to
-           the specified value.";
+          "Value and comparison operations for conditions based on
+           the value of the BGP LOCAL_PREF.";
+        reference
+          "RFC 4271: Section 5.1.5.";
+
+        leaf value {
+          type uint32;
+          description
+            "BGP LOCAL_PREF value for comparison to the entry in the
+             BGP route.";
+        }
+
+        choice operation {
+          case eq {
+            leaf eq {
+              type empty;
+              description
+                "Check to see if the value is equal.";
+            }
+          }
+
+          case lt-or-eq {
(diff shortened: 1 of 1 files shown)


src/yang/ietf-bgp-policy.yang
```diff
@@ -148,6 +148,38 @@ module ietf-bgp-policy {
        and setting it to the IGP cost (predefined value).";
   }
 
+  grouping equality-operator {
+    description "";
```
mjethanandani, line 152: Did you mean to put a description here?

src/yang/ietf-bgp-policy.yang
```diff
@@ -292,11 +324,39 @@ module ietf-bgp-policy {
...
-      leaf med-eq {
-        type uint32;
+      container local-pref {
         description
-          "Condition to check if the received MED value is equal to
-           the specified value.";
+          "Value and comparison operations for conditions based on
+           the value of the BGP LOCAL_PREF.";
+        reference
+          "RFC 4271: Section 5.1.5.";
```
mjethanandani, line 332: Nit picking again, but you need the RFC title here.

src/yang/ietf-bgp-policy.yang
```diff
@@ -292,11 +324,39 @@ module ietf-bgp-policy {
...
+        uses equality-operator;
+      }
+
+
+      container med {
+        description
+          "Value and comparison operations for conditions based on
+           the value of the BGP MULTI_EXIT_DISC (MED)";
+        reference
+          "RFC 4271: Section 5.1.4.";
```
mjethanandani, line 350: Same here.

mjethanandani (commented):
Do all the operations apply for each of the values? For example, we had med-eq because I thought that either the value matched or it did not. What happens if the med value is not equal?

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: This keeps happening because I continue to copy and paste from elsewhere.  Please take up the normalization Issue as I work through the rest of the document changes.

mjethanandani:
> > Do all the operations apply for each of the values? For example, we had med-eq because I thought that either the value matched or it did not. What happens if the med value is not equal?
> 
> You'll note that you can still do "med equal" in this form.

Ok. Maybe I did not make myself clear. In the operation
> 
> Looking at the diff this morning, we may also want something supporting a range operation. E.g., between 100-200. This would require the ability to join the inequalities with an AND operation.


> > Do all the operations apply for each of the values? For example, we had med-eq because I thought that either the value matched or it did not. What happens if the med value is not equal?
> 
> You'll note that you can still do "med equal" in this form.

Ok. Maybe I did not make myself clear. Instead of 'eq', if the user were to configure 'lt-or-eq' for med, would that still be a valid configuration?


> 
> Looking at the diff this morning, we may also want something supporting a range operation. E.g., between 100-200. This would require the ability to join the inequalities with an AND operation.

src/yang/ietf-bgp-policy.yang
suehares: Mahesh: 
Do you need RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 5.1.5? 

Just curious.

suehares:
Did the range on "it-or-eq" get resolved.  I do not find the resolution in this thread.

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: Resolved in next patch.

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: Resolved in next patch.

src/yang/ietf-bgp-policy.yang
suehares: Yes - I found the resolution in Issue #26.  He meant RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 5.1.5.

suehares:
Here's what I think needs to be resolved to close this pull request: 
1. "it-or-eq" issue resolved (see above)  - answer from Jeff or Mahesh
2. Conflict in branches - conflict Jeff or Mahesh

jhaas-pfrc:
> Ok. Maybe I did not make myself clear. Instead of 'eq', if the user were to configure 'lt-or-eq' for med, would that still be a valid configuration?

For match conditions, yes.  Cisco has it.  As noted in the other PR that was merged, Juniper would like it.

> I am now reading your comments in the other PR for inequality comparisons that Cisco supports, and if that is true, my question above is answered. What that also means is that if Juniper was to keep status-quo, they would minimally have to deviate the model and remove the unequal operations. Right?

Correct.  The comments from that PR are the most applicable here.  We really don't want implementors to have to augment in logic in conditions if we can help it since there's a strong need for consistency.  Deviation is fine here, I think.

We'll see how this holds up for IETF review.

If it doesn't, we'll likely end up with a set of groupings covering different classes of match comparisons for equality and inequality, potentially with negation.

suehares (approved):

mjethanandani:
> I believe this resolves Mahesh lt-eq. Please comment if it does not.

It does.

mjethanandani (approved):

PR #288, opened 2023-01-06
Issue #242 - Remove no-client reflect.

### LINKED ISSUE #242: route-reflector/no-client-reflect depends on route-reflector/client
The structure of no-client-reflect has it as a peer-node with client.  Typically implementations will make no-client-reflect a child of the client node.

A few thoughts:
Option 1: Restructure client into a container rather than a leaf, and make it presence?  This would let us make no-client-reflect live in the hierarchy appropriately.
Option 2: Create a "must" dependency between the two leaves.

jhaas-pfrc:
Deployed implementations of BGP route-reflectors are highly inconsistent about their support for "no-client-reflect".  Since the set of if-features necessary to represent the whole set of deployed behaviors would unnecessarily complicate the model, remove the feature for the time being.

Closes #242

### DIFF
--- a/src/yang/ietf-bgp-common-structure.yang
+++ b/src/yang/ietf-bgp-common-structure.yang
@@ -105,15 +105,6 @@ submodule ietf-bgp-common-structure {
           "RFC 4456: BGP Route Reflection: An Alternative to
                      Full Mesh.";
       }
-      leaf no-client-reflect {
-        type boolean;
-        default "false";
-        description
-          "When set to 'true', this disables route redistribution
-           by the Route Reflector. It is set 'true' when the client
-           is fully meshed in its peer-group to prevent sending of
-           redundant route advertisements.";
-      }
       leaf client {
         type boolean;
         default "false";


mjethanandani (commented):
I would prefer that we add in a feature statement for no-client-reflect, and keep the leaf.

But I will let Sue and Keyur comment on it.

jhaas-pfrc:
> I would prefer that we add in a feature statement for no-client-reflect, and keep the leaf.
> 
> But I will let Sue and Keyur comment on it.

If you'd prefer to try to keep it, you have the discussion on what the behavior should be like in each implementation.  Please submit modeling for how you think the feature should work as a different pull request.

jhaas-pfrc:
For now, close this pull request in the hopes that Mahesh finds a different way to resolve the issue.  It can be re-opened if we decide to proceed with ripping it out.

mjethanandani:
In discussion with Keyur and Derek, we believe the best option is accept this PR.

mjethanandani (approved):

PR #295, opened 2023-01-12
Issue #294 - Remove graceful-restart operational state from peer-groups

### LINKED ISSUE #294: Graceful restart op state doesn't belong in bgp/peer-groups
grouping neighbor-and-peer-group-common needs to have the op state for container graceful-restart removed from it in the peer-group case.  This will involve refactoring graceful restart from the grouping.

jhaas-pfrc:
Closes #294

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -252,6 +252,8 @@ module ietf-bgp {
           description
             "Parameters relating the graceful restart mechanism for
              BGP.";
+          reference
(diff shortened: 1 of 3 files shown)


src/yang/ietf-bgp.yang
```diff
@@ -448,6 +384,70 @@ module ietf-bgp {
           }
 
           uses neighbor-group-config;
+          container graceful-restart {
```
mjethanandani, line 387: Should we add a reference to RFC 4724 somewhere here?

mjethanandani (approved):

PR #299, opened 2023-01-12
Issue #298 - Make treat-as-withdraw group and neighbor common.

### LINKED ISSUE #298: treat-as-withdraw missing at peer-groups level
The treat-as-withdraw configuration was only in bgp/neighbors and needs also to be in bgp/peer-groups

jhaas-pfrc:
The treat-as-withdraw configuration was only in bgp/neighbors and needs also to be in bgp/peer-groups

Closes #298

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -220,7 +220,17 @@ module ietf-bgp {
       }
     }
 
-    uses structure-neighbor-group-error-options;
+    leaf treat-as-withdraw {
+      type boolean;
+      default "false";
+      description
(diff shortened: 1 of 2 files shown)


src/yang/ietf-bgp.yang
```diff
@@ -219,6 +219,8 @@ module ietf-bgp {
            graceful restart with the peer";
       }
     }
+
+    uses structure-neighbor-group-error-options;
```
mjethanandani, line 223: I am not a big fan of single use grouping. If this is the only place this grouping is going to be used, I would rather that we inline the grouping.

mjethanandani (approved):

Issue #292, opened 2023-01-10
OC vs. IETF model high level content variations of interest


jhaas-pfrc:
Note that this comparison is done primarily in the direction of "what is in the OC model that is interesting that may not be in the IETF model".  A few comparative items in the IETF model vs. OC for "should this be present?" are also listed.

IETF-only:
enable-med
med-plus-igp
peer-groups/graceful-restart peer op state - likely a bug

OC-only:
afis-safis/extended-next-hop-encoding
dynamic-neighbor-prefixes
dynamically-configured leaf
disable-peer-as-filter
neighbor/afi-safis/state/prefixes/received-pre-policy
neighbor/afi-safis/use-multiple-paths - IETF only has in afi-safis
peer-groups/state/total-paths/prefixes
peer-groups/treat-as-withdraw
peer-groups/afi-safis/add-paths
rib/tunnel-encapsulation
sr-te-policy specific items
evpn specific items
policy/neighbor-sets

Differences:
IETF-model only does add-paths at neighbors level.  OC is at global afi-safi level.

IETF-model has apply-policy at global/afi-safi, global, neighbor, neighbor/afi-safi, peer-groups, peer-groups/afi-safi. OC has neighbors, neighbors/afi-safis, peer-groups, peer-groups/afi-safis

prefix-limit in IETF has max-prefixes, shutdown-threshold-pct, restart-timer; OC has max-prefixes, prevent-teardown, warning-threshold-pct.

OC has peer-group at top of neighbor config
OC has neighbor local-address,  (neighbor-port),  read-write
OC using old style "auth-password"
OC has only enable for damping, no parameters
OC has graceful-restart restart-time in timers container rather than GR container.
OC BFD is only enable-bfd
OC does policy comparators via identity rather than choice
OC has differentiated the set-community "method", very different structure

Needs additional scrutiny:
communities, extended communities
policy set options

jhaas-pfrc:
I have rather mixed feelings about incorporating the  neighbor/afi-safis/state/prefixes/received-pre-policy state since I don't think that's good generic BGP modeling.  It might be appropriate to some platforms though.

In general, the prefix counts in the adj-ribs aren't expected to change depending on the logical view.  If you receive 10 routes in your rib-in, retain 10 routes (else, how can you count them?), but only 5 are eligible to move into loc-rib, there's still 10 entries in rib-in, but may be 5 in loc-rib if those are the selected item.

[...]

PR #301, opened 2023-01-12
Issue #300 - add-paths is missing at afi-safis level configuration

### LINKED ISSUE #300: add-paths is missing at afi-safis level configuration
Noted as art of review in Issue #292

jhaas-pfrc:
Closes #300

### DIFF
--- a/src/yang/ietf-bgp-common-structure.yang
+++ b/src/yang/ietf-bgp-common-structure.yang
@@ -153,10 +153,10 @@ submodule ietf-bgp-common-structure {
     }
   }
 
-  grouping structure-neighbor-group-add-paths {
+  grouping structure-add-paths {
     description
       "Structural grouping used to include ADD-PATHs configuration
-       and state for both BGP neighbors and peer groups";
+       and state.";
     container add-paths {
       if-feature "bt:add-paths";
       description
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -338,6 +338,7 @@ module ietf-bgp {
               uses mp-afi-safi-graceful-restart-config;
             }
             uses route-selection-options;
+            uses structure-neighbor-group-add-paths;
             uses global-group-use-multiple-paths;
             uses mp-all-afi-safi-list-contents;
           }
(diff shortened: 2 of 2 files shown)


jhaas-pfrc:
Partially addresses Issue #292

mjethanandani (commented):
This is more for my education.

The comment in the grouping says that it is used to include ADD-PATHs configuration for neighbors and peer-groups. As I see the change, the first addition applies it at a global level. Is this applicable at a global level? If so, should the description of the grouping reflect that?

jhaas-pfrc:
> This is more for my education.
> 
> The comment in the grouping says that it is used to include ADD-PATHs configuration for neighbors and peer-groups. As I see the change, the first addition applies it at a global level. Is this applicable at a global level? If so, should the description of the grouping reflect that?

Good point.  Let me ponder if the better answer is to update grouping name/description or whether further work needs to bedone.

jhaas-pfrc:
> This is more for my education.
> 
> The comment in the grouping says that it is used to include ADD-PATHs configuration for neighbors and peer-groups. As I see the change, the first addition applies it at a global level. Is this applicable at a global level? If so, should the description of the grouping reflect that?

So, global afi/safi support for add-paths is in the OC model, partially because of Cisco support:
https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus3548/sw/unicast/6x/b_Cisco_Nexus_3548_Unicast_Routing_Configuration_Guide_Release_6_x/b_Cisco_Nexus_3548_Unicast_Routing_Configuration_Guide_Release_6_x_chapter_0111.pdf

Juniper only supports it at the group and neighbor level:
https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/add-path-edit-protocols-bgp.html

Arista has support for global add-paths as well.
https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp

jhaas-pfrc:
I'm making the call that global afi/safi will be included.  Juniper will have to deviate.

Please review the name change and if we're good, we can merge.

Partially addresses Issue #292

mjethanandani (approved):

suehares:
https://forum.huawei.com/enterprise/en/multiple-paths-for-the-same-prefix-to-be-advertised-in-bgp/thread/592456-861

Shows an example of Huawei at AFI/SAFI level.  This adds this input to the trouble ticket.    it does not need to be opened again.

PR #315, opened 2023-01-18
Issue #262 move as-path-sets to top of matches

### LINKED ISSUE #262: Move as-path-asets in routing-policy to be prior to communities
Note that the contents of as-path-sets may change as part of doing the regex work that is pending.

In general, look at the output in policy and sort it in a more pleasant manner.

```
   module: ietf-bgp-policy

     augment /rt-pol:routing-policy/rt-pol:defined-sets:
       +--rw bgp-defined-sets
          +--rw community-sets
          |  +--rw community-set* [name]
          |     +--rw name      string
          |     +--rw member*   union
          +--rw ext-community-sets
          |  +--rw ext-community-set* [name]
          |     +--rw name      string
          |     +--rw member*   union
          +--rw large-community-sets
          |  +--rw large-community-set* [name]
          |     +--rw name      string
          |     +--rw member*   union
          +--rw as-path-sets
          |  +--rw as-path-set* [name]
          |     +--rw name      string
          |     +--rw member*   string

```

jhaas-pfrc:
Aesthetic cleanup.

Closes #262

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -191,6 +191,33 @@ module ietf-bgp-policy {
     container bgp-defined-sets {
       description
         "BGP-related set definitions for policy match conditions.";
+      container as-path-sets {
+        description
+          "Enclosing container for list of define AS path sets.";
+        list as-path-set {
+          key "name";
+          description
+            "List of defined AS path sets.";
+          leaf name {
+            type string;
+            description
+              "Name of the AS path set -- this is used to reference
+               the set in match conditions.";
+          }
+          leaf-list member {
+            type string;
+            description
+              "AS path regular expression using BGP YANG AS_PATH
+               regular expression syntax.  If any of the regular
+               expressions in the lists are matched, the as-path-set
+               is considered matched.";
+            reference
+              "RFC XXXX: BGP YANG Model for Service Provider
+               Networks, Appendix F.3.";
+          }
+        }
+      }
+
       container community-sets {
         description
           "Enclosing container for list of defined BGP community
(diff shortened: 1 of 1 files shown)


src/yang/ietf-bgp-policy.yang
```diff
@@ -191,6 +191,33 @@ module ietf-bgp-policy {
...
+          leaf name {
+            type string;
+            description
+              "Name of the AS path set -- this is used to reference
+               the set in match conditions.";
+          }
+          leaf-list member {
+            type string;
+            description
+              "AS path regular expression using BGP YANG AS_PATH
```
mjethanandani, line 210: The description says 'member' is a string with a AS path regular expression. But there is no pattern statement here to reflect the AS path regular expression. I see that we have a reference to the new section on AS PATH regular expression, but that is informative. Should this explain (or have an issue) to track if/when we do define a regular expression.

src/yang/ietf-bgp-policy.yang
```diff
@@ -191,6 +191,33 @@ module ietf-bgp-policy {
...
+               regular expression syntax.  If any of the regular
+               expressions in the lists are matched, the as-path-set
+               is considered matched.";
+            reference
+              "RFC XXXX: BGP YANG Model for Service Provider
```
mjethanandani, line 215: Please update the reference to the latest title of the draft.

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: As we had discussed, it makes zero sense to try to list the rules for how to create a regular expression in general, much less the subset we're creating for such a language element, in the leaf.  So, we're not going to have anything here.

I'm rusty on my compiler theory, but it's my memory that you can't express the output of fully arbitrary balanced parenthesis matches as a regular expression.  I think you need a CFG for that.  Thus, I don't think we can use a YANG leaf pattern to help here.

src/yang/ietf-bgp-policy.yang
jhaas-pfrc: If you're going to insist on me cleaning up every bit of dangling reference and punctuation cleanup whenever I try to do structural cleanup, I'm happy to stop doing this work until the rest of the "authors" do the tasks assigned in the other cleanup issues.

jhaas-pfrc: Addressed in other PR. Thanks.

PR #316, opened 2023-01-18
Issue #307 Reconcile prefix-limit vs. OpenConfig model

### LINKED ISSUE #307: Reconcile prefix-limit vs. OC model
The IETF model started from OC.  OC has since changed the semantics of their prefix-limit behaviors in the prefix-limit container.  Similarly, there's a neighbor leaf last-prefix-limit-exceeded that provides information about when things last reset.

Review both sets of semantics against implementations and reconcile them.

jhaas-pfrc:
- Change shutdown-threshold-pct to warning-threshold-pct to be consistent with OC.  It's more consistent with the various implementations of prefix-limits as well.
- Change restart-timer to idle-time.  This is a change contrary to OC, but is motivated by implementations that want to permit the peers to be idled forever when the prefix-limit trips.  The name change better suits the semantic of the time to wait in idle, which could be "forever".
- Add a "teardown" leaf.  This is contrary to OC which uses a "prevent-teardown" leaf.  Surveyed implementations in some cases don't have the option to teardown, and a default of false for "prevent-teardown" means implementations that want to deviate the model need to radically change the content in order to say that they can't do so.  With "teardown" as a positive action, it's clear that deviating the leaf away means that teardown doesn't need to be supported.
- Add a new per-neighbor leaf, prefix-limit-restart-time to indicate when the peer session will restart.  Largely verbatim from OC.
- Made prefix-limit available at the peer and peer-group level.  Multiple implementations only do per-neighbor level prefix-limits.

Structural cleanup also done in this diff:
- Remove the grouping neighbor-group-timers-config, which was only used once.
- Moved common prefix-limit elements into a -config and -state grouping to permit re-use in neighbors and peer-groups containers.

Closes #307
Partially addresses #292


suehares (approved):

suehares:
Changes seem reasonable.  Approved.

PR #319, opened 2023-01-19
Issue #281 Eliminate grouping neighbor-group-config

### LINKED ISSUE #281: Eliminate grouping neighbor-group-config ?
Currently, grouping neighbor-group-config is used in exactly the same places that grouping neighbor-and-peer-group-common are used, so they're really the same thing.

After doing the comparative analysis vs. the OpenConfig model, consider removing this group if nothing gets added to differentiate things.

jhaas-pfrc:
grouping neighbor-group-config is used in exactly the same places that grouping neighbor-and-peer-group-common are used, so they're really the same thing.

Eliminate neighbor-group config.  This required restructuring some of the dependencies of the submodules.

Note that this causes the graceful-restart config and state to bubble down in the hierarchy where used in the neighbors and peer-groups.

Closes #281

### DIFF
--- a/src/yang/ietf-bgp-common-structure.yang
+++ b/src/yang/ietf-bgp-common-structure.yang
@@ -14,8 +14,6 @@ submodule ietf-bgp-common-structure {
     reference
       "RFC XXXX: YANG Model for Border Gateway Protocol (BGP-4).";
   }
-  include ietf-bgp-common-multiprotocol;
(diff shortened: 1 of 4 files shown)


suehares:
AFAIK - I was not requested to review or approve this pull.

keyur-arrcus (approved):
Looks good.

PR #321, opened 2023-01-19
Issue #148 RFC 8538 encapsulated notification support

### LINKED ISSUE #148: Support for RFC 8538
For after 1.0.

Impacts the graceful restart flags and adds to notification message types.

jhaas-pfrc:
Since a bit identity had to be registered for the Graceful restart type and it was better to simply add the RFC 8538 bit as part of our initial version, finish the work to display encapsulated notifications from RFC 8538 when processing a CEASE/HARD RESET.

Since these leaves are optional, there's no need to protect it with an if-feature.

Closes #148

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -244,6 +244,55 @@ module ietf-bgp {
     }
   }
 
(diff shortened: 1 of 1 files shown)


suehares (approved):
Good approach.

PR #323, opened 2023-01-20
Issue #309 Dynamic peers feature

### LINKED ISSUE #309: Dynamic BGP neighbors
As part of reviewing the differences in the IETF model and the OC model in Issue #292, it was noted that the IETF model doesn't currently have anything that covers dynamic neighbors. 

Several implementations have a form of this configuration. E.g., "allow" in Juniper's implementation and GateD descendants.

Review vs. current implementations and suggest appropriate modeling.  This may be optional, so tagging this as enhancement in case there's insufficient commonality to warrant an IETF feature in initial model release.

jhaas-pfrc:
Many implementations of BGP support dynamic peers.  Most tend to configure it from peer-group scope.  This commit adds dynamic peers as a grouping that is used at the per-group scope.  Its structure is a container with a single list where the prefixes of the peers are configured.

While this modeling is mostly empty space, it enables two flavors of extensibility:
1. The list keyed by prefix can have additional per-prefix configuration.  This accommodates features in some vendors to further restrict peers on a per-prefix basis.
2. The empty container allows per-container augmentation as well.

If an implementation prefers to model this similar to OpenConfig with a top-level bgp dynamic peers container, this grouping can be augmented in at the global level to provide naming consistency.

Closes #309

### DIFF
--- a/src/yang/ietf-bgp-common-structure.yang
+++ b/src/yang/ietf-bgp-common-structure.yang
@@ -14,6 +14,11 @@ submodule ietf-bgp-common-structure {
     reference
       "RFC XXXX: YANG Model for Border Gateway Protocol (BGP-4).";
   }
+  import ietf-inet-types {
+    prefix inet;
+    reference
+      "RFC 6991: Common YANG Data Types.";
+  }
   include ietf-bgp-common-multiprotocol;
   include ietf-bgp-common;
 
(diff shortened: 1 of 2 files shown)


suehares (approved):
Dynamic peers are a useful addition to the model.   I agree with the addition of this code.

PR #325, opened 2023-01-20
Issue #225 A typedef for BGP color extended communities

### LINKED ISSUE #225: Add support for color based ext-community-type.
The current does not support color based bgp-ext-community-type as defined by RFC 9012. While the OC model does support it using this string definition, there seems to be a bug.

```
RFC 9012:
4.3. Color Extended Community
The Color Extended Community is a Transitive Opaque Extended Community with the encoding shown in Figure 15.

   0                   1                   2                   3
   0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  | 0x03 (1 octet)| 0x0b (1 octet)|        Flags (2 octets)       |
  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  |                      Color Value (4 octets)                   |
  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

...
14.10. Color Extended Community Flags
IANA has created a registry named "Color Extended Community Flags" under the "Border Gateway Protocol (BGP) Tunnel Encapsulation" grouping [IANA-BGP-TUNNEL-ENCAP]. The registration policy for this registry is "Standards Action". The minimum possible value is 0, and the maximum is 15.

This new registry contains columns for "Bit Position", "Description", and "Reference". No values have currently been registered.
```

The OC model defines the pattern for it as follows:

```
        type string {
        // Extended Color Community
        pattern '^color:'                                                    +
                '[0-1]{2}:'                                                  +
                '(429496729[0-5]|42949672[0-8][0-9]|4294967[0-1][0-9]{2}'    +
                '|429496[0-6][0-9]{3}|42949[0-5][0-9]{4}|4294[0-8][0-9]{5}|' +
                '429[0-3][0-9]{6}|4[0-1][0-9]{7}|[1-3][0-9]{9}|'             +
                '[1-9][0-9]{1,8}|[0-9])$';
      }
```

Removing the ^ and $, the flags field seems to allow for 00, 01, 10, 11 only for the flags field, which is probably not what the RFC intent is.

jhaas-pfrc:
I'm not sure we want to proceed with this commit in the initial IETF BGP YANG module, but this is a useful modeling exercise for how to solve some ugly issues.

The BGP Color community has changed in RFC 9012 from its older format to move the first 16 bits of RESERVED value into a Flags field.  SR-TE policy work has allocated two of those bits.  A number of bits are currently unknown and may be assigned further in the future.

A flexible format where, when the flags are all known, the flag fields are displayed as a comma-separated set of flag bits.  If any flag bits are unknown, or the Flags are zero, a default format to render the Flags as a uint16 is done.

Note that this varies from the OC proposal to simply display the two known flags as it doesn't cover future flag possibilities.

Closes #225

### DIFF
--- a/src/yang/iana-bgp-types.yang
+++ b/src/yang/iana-bgp-types.yang
@@ -771,6 +771,38 @@ module iana-bgp-types {
       }
 
       type string {
+        pattern 'color:'
+              + '(((co=[01][01],?):)|'
+              + '((6[0-5][0-5][0-3][0-5]|[1-5][0-9]{4}|'
+              + '[1-9][0-9]{1,4}|[0-9]):))'
+              + '(4[0-2][0-9][0-4][0-9][0-6][0-7][0-2][0-9][0-6]|'
+              + '[1-3][0-9]{9}|[1-9]([0-9]{1,7})?[0-9]|[0-9])';
+        /*
+         * description
+         *   "Type 0x03, Sub-Type 0x0b: Color
+         *    color:(16-bit decimalized flags):(color)
+         *    color:co=(flag bits):(color) - color-only
+         *    16-bit decimalized flags [01][01]
+         *    4 octets color.
+         *
+         *    When the Flags field is zero, or there are unknown Flag
+         *    values, the flags are rendered as a decimal value.
+         *    When all flags are known, they are rendered with their
+         *    flag name (e.g. co) as comma-separated sets.  The final
+         *    flag SHOULD NOT include a trailing comma.  The order of
+         *    flag names MUST be in the order of the assigned bits to
+         *    enable matching via regular expressions.
+         *
+         *    New flags may be allocated in futured RFCs.";
+         * reference
+         *   "RFC 9012: The BGP Tunnel Encapsulation Attribute,
+         *    Section 4.3,
+         *    draft-ietf-idr-segment-routing-te-policy: Advertising
+         *    Segment Routing Policies in BGP, Section 3.";
+         */
+      }
+
+      type string {
         // raw with 8 octets
         pattern 'raw:([0-9A-F][0-9A-F]:){7}[0-9A-F][0-9A-F]';
       }


suehares:
Jeff: 
I agree with leaving the typedef for Color Community until another time. 

I did not understand this sentence: " If any flag bits are unknown, or the Flags are zero, a default format to render the Flags as a uint16 is done."

If you have the two known flag bits set, would it be "flag1, flag2".  
If you have all know known flag bits zero,  would it be "0" or "none"?

suehares (changes requested):
I request we leave this for a later model or an augmentation (for color) of the BGP model. 
Therefore, I am asking you not to merge in the changes. 

I agree with Jeff's proposal.

jhaas-pfrc:
> Jeff: I agree with leaving the typedef for Color Community until another time.
> 
> I did not understand this sentence: " If any flag bits are unknown, or the Flags are zero, a default format to render the Flags as a uint16 is done."
> 
> If you have the two known flag bits set, would it be "flag1, flag2". If you have all know known flag bits zero, would it be "0" or "none"?

Consider the case we have right now.  RESERVED was partitioned into color-only for bits 0,1 and renamed the uint16 as Flags.

If the value of the Flags is zero, we omit it in the output for consistency with older spec.  So, color type/subtype flags=0, color=100 renders as "color:100".
If only the color-only bits are set, we render them.  For example, color:co=01:100
Let's say the value was something outside the known color-only flags and the uint16 is value 100.  We then render in the non-bit-flag format: color:100:100

In this mechanism, when we add known bits, we start adding the names.  So, if bit 2 is given the name foo, and bit 3 is bar, and co=01 and foo and bar are set with color 100, the rendered format would be color:co=01,foo,bar:100

the requirement is that the known flag bits are canonically ordered such that regular expressions can test for them.  Consider policy where you want to select color communities with color value 100 and foo set and everything else is irrelevant.  The regex would be "^color:.*foo.*:100$"

keyur-arrcus (commented):
Looks good. Ok by me to leave it out for a while as Jeff and Sue discussed.

suehares (approved):

mjethanandani:
Not merging this PR as suggested by Jeff and Sue.

PR #328, opened 2023-01-24
Issue #327 Split the BGP Community IANA types into a separate module.

### LINKED ISSUE #327: Split BGP community types into separate IANA maintained module
Per discussion, since the BGP community types will get maintained at significantly greater frequency than the rest of the registered BGP types, split the module.

jhaas-pfrc:
Split the BGP community related IANA types into a separate module. This will permit these items which will churn more often to be in a separate file.

Closes #327


keyur-arrcus (approved):
Looks good. Few comments -

Would be good to clean up and have consistent format for comments.

This file covers all communities - Standard, extended and large. I am assuming in future Wide will belong here as well?

We have route-origin and route-target extended community. Does IPv6 Address specific Ext comm, 4 octet AS specific, Source AS, and other extended communities belong here as well?
https://www.iana.org/assignments/bgp-extended-communities/bgp-extended-communities.xhtml

jhaas-pfrc:
@keyur-arrcus 
> > Would be good to clean up and have consistent format for comments.

Before I do the commit, please give an example of inconsistencies you're seeing.

> This file covers all communities - Standard, extended and large. I am assuming in future Wide will belong here as well?

I suspect Wide communities will want its own module.  Much like flowspec and bgp-ls, it's going to be an interesting modeling challenge.

> We have route-origin and route-target extended community. Does IPv6 Address specific Ext comm, 4 octet AS specific, Source AS, and other extended communities belong here as well? https://www.iana.org/assignments/bgp-extended-communities/bgp-extended-communities.xhtml

I'll restate that as "support for RFC 5701".  I'll open an issue for it, but we should consider that a different issue rather than tack that onto here.

suehares (approved):

jhaas-pfrc:
Keyur notes that this is about the comment style.  I have chosen C-style /* for multi-line and // for single line.
Will verify with @mjethanandani that this is consistent with yang doctor practices.

PR #341, opened 2023-04-03
route-reflector/no-client-reflect depends on route-reflector/client

mjethanandani:
Address issue #242


mjethanandani:
@jhaas-pfrc, contrary to what I we discussed, my understanding of presence statement was correct.

In RFC 7950, it says:

   If a container has the "presence" statement, the container's
   existence in the data tree carries some meaning.

And as an example RFC 7950 gives this:

```
     container system {
       description
         "Contains various system parameters.";
       container services {
         description
           "Configure externally available services.";
         container "ssh" {
           presence "Enables SSH";
           description
             "SSH service-specific configuration.";
           // more leafs, containers, and stuff here...
         }
       }
     }
```
   A corresponding XML instance example:

```
     <system>
       <services>
         <ssh/>
       </services>
     </system>
```
   Since the <ssh> element is present, SSH is enabled.

suehares (approved):
Look good.  I have not check anything with the tools.

mjethanandani:
After discussing with Keyur and Derek, we feel that it is best to accept the remove of no-client-reflect PR that @jhaas-pfrc had proposed in January, and close this PR out.

PR #350, opened 2023-04-21
Updated the WG Web URL.

mjethanandani:
From tools.ietf.org to datatracker.ietf.org


suehares (approved):
Discussion on 4/21 at meeting

PR #360, opened 2023-07-26
Add RD in the BGP global configuration.

mjethanandani:


mjethanandani:
RD needs to be defined under a NI, and can be used outside of BGP also, e.g. EVPN for MAC learning without BGP. As such this should be considered as part of a new draft that extends the NI model to add not only RD but also RT.

PR #376, opened 2025-12-09
Remove YANG modules that have been published as part of a RFC.

### LINKED ISSUE #372: Start using dependent YANG modules in IANA
Quite a few of the YANG models that this draft imports are now RFCs. Rather than extract them from I-D, we should move to downloading the modules directly from IANA.

mjethanandani:
Address issue #372

### DIFF
--- a/src/download-dependent-models.py
+++ b/src/download-dependent-models.py
@@ -1,14 +1,9 @@
 import os
 
 list_of_ietf_models =\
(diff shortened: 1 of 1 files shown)


jhaas-pfrc (approved):

suehares:
Works for me.

suehares (approved):

PR #393, opened 2026-01-18
Issue 392: leaf session-state should be config false

### LINKED ISSUE #392: leaf session-state was missing config false
There was commented-out text in the session-state leaf:

```
            //  notification does not like a non-config statement.
            //  config false;
```

Removing the comment seems to build fine?

jhaas-pfrc:
Fixes: #392

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -612,8 +612,7 @@ module ietf-bgp {
                    established.";
               }
             }
-            //  notification does not like a non-config statement.
-            //  config false;
+            config false;
             description
               "The BGP peer connection state.";
             reference


src/yang/ietf-bgp.yang
```diff
@@ -612,8 +612,7 @@ module ietf-bgp {
                    established.";
               }
             }
-            //  notification does not like a non-config statement.
-            //  config false;
+            config false;
```
mjethanandani, line 615: I remember encountering an error when I had the 'config false' statement. It was not clear to me why pyang (or was it yanglint) was complaining about it. Are you saying you do not see an error now?

jhaas-pfrc:
@mjethanandani will verify that this error is fixed during build to make sure I'm not missing something.

mjethanandani (commented):
I checked and no longer see the error.

mjethanandani (approved):

PR #410, opened 2026-01-20
Issue 401: Add values to as path segment identities.

### LINKED ISSUE #401: Identities for as path segment types should include value
For example, the entry below should say the type has a value of 1.  See example in other identities that have enumerated values.

```
  identity as-set {
    base as-path-segment-type;
    description
      "Unordered set of autonomous systems that a route in the UPDATE
       message has traversed.";
    reference
      "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 4.3.";
  }
```

jhaas-pfrc:
Fixes: #401

### DIFF
--- a/src/yang/iana-bgp-types.yang
+++ b/src/yang/iana-bgp-types.yang
@@ -168,7 +168,8 @@ module iana-bgp-types {
     base as-path-segment-type;
     description
       "Unordered set of autonomous systems that a route in the UPDATE
-       message has traversed.";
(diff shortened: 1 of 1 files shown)


mjethanandani (approved):
I assume that these values are for informational purposes, as identities do not carry a value in themselves.

suehares (approved):
Thanks for adding in the values. Good to go.

PR #421, opened 2026-01-24
Issue 416: Move as-path-segment-type to an enum

### LINKED ISSUE #416: Migrate segment types to enum.
In discussing this with @mjethanandani, the type of an identity is a weirdness.  Presume we get a new type-5 path segment that didn't previously exist.  With identities, there's no way to display this.  With an enum, we can simply display the value "5" rather than nothing.

jhaas-pfrc:
Fixes: #416

### DIFF
--- a/src/yang/iana-bgp-types.yang
+++ b/src/yang/iana-bgp-types.yang
@@ -154,48 +154,47 @@ module iana-bgp-types {
 
   /* BGP AS_PATH Segment Type Identities. */
 
-  identity as-path-segment-type {
-    description
-      "Base AS Path Segment Type. In [BGP-4], the path segment type
-       is a 1-octet field with the following values defined.";
-    reference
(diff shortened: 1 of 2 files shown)


suehares (commented):
I believe the the reason you are doing an ENUM is to get the value according to value set in the BGP specification.  Is this correct?

suehares (approved):
I approve of the approach of the enum rather than the identity.

jhaas-pfrc:
> I believe the the reason you are doing an ENUM is to get the value according to value set in the BGP specification. Is this correct?

There are three reasons:
1. If we use an identity, it's good to list the code point in the descriptive text.  With enums, it's explicit
2. The YANG model maintenance considerations that encourage the use of identities don't apply here.  This will be singly managed as a single namespace and coordinated only through IANA and IETF process.  Thus we don't have to worry about maintenance collisions.
3. enums do better with unknown values.  Without it, it means some future as-path segment type that an older implementation receives couldn't be displayed without a module update.

PR #425, opened 2026-01-24
Issue 424: Avoid wasting rsync work for yang-paramters

### LINKED ISSUE #424: Avoid doing repeated rsync's of yang-parameters
The yang-parameters are downloaded using rsync to the container's bin directory.
That bin directory is destroyed as part of make clean.
Destroying the directory wastes the benefits of rsync to avoid work.

Proposed fix:
move yang-parameters to the container's root
symbolically link in that directory to bin so the rest of the build infra picks it up.

jhaas-pfrc:
Fixes: #424

### DIFF
--- a/src/validate-and-gen-trees.sh
+++ b/src/validate-and-gen-trees.sh
@@ -3,8 +3,8 @@
 #
 # Does the user have all the IETF published models.
 #
-if [ ! -d ../bin/yang-parameters ]; then
-   rsync -avz --delete rsync.iana.org::assignments/yang-parameters ../bin/
+if [ ! -d ../yang-parameters ]; then
+   rsync -avz --delete rsync.iana.org::assignments/yang-parameters ../
 fi
 
 for i in ../bin/ietf-*\@$(date +%Y-%m-%d).yang
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ src/yang/example-l2vpn-configuration-*
 src/yang/ietf-bgp.mib*
 src/yang/show-*
 state/*
+yang-parameters
 README.ncs
 ncs.conf
 setup.mk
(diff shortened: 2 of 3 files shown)


suehares (approved):
Thanks for catching this Jeff.

src/validate-and-gen-trees.sh
```diff
@@ -3,8 +3,8 @@
 #
 # Does the user have all the IETF published models.
 #
-if [ ! -d ../bin/yang-parameters ]; then
-   rsync -avz --delete rsync.iana.org::assignments/yang-parameters ../bin/
+if [ ! -d ../yang-parameters ]; then
+   rsync -avz --delete rsync.iana.org::assignments/yang-parameters ../
```
mjethanandani, line 7: I am not sure if this change is needed.

The reason why rsync runs everytime is because of this line in start.sh

make clean

which in turn runs the following in draft folder

make -C draft clean

which is the one that removes the bin directory and forces rsync to download the YANG modules.

In other words, if you do not want to have rsync download everytime, just remove the 'make clean' in start.sh or if you want a more surgical fix, remove the following line in draft/Makefile

-rm -rf ../bin

Having siad that, I wish there was a way to know that one of the YANG modules has changed on the remote site for one to trigger a download. Otherwise, I find myself dealing with old modules.

mjethanandani (changes requested):

src/validate-and-gen-trees.sh
jhaas-pfrc: As someone who has now lost hours of their life in the last few weeks to avoiding the negative impact of both rsync and previously idnits wanting to tickle the network for things that will never change - I rather disagree that "it's not needed".  I've been running on a local branch with those two diffs applied for the last week-plus to get the benefit of the faster builds.

What I won't necessarily make a case for is "my fix is the right one", but I think this pattern should be addressed because your infra is what we're encouraging people to use.

Local hacks to the makefile to avoid doing work repeatedly means that you have something you have to stash each time you do work on a branch or certain types of rebases.  It's a non-helpful pattern.

The files in question are grabbed via rsync.  Letting them be cached in the filesystem between builds is quite reasonable.  But, to do so, they need a home. 

Your pattern for bin is really a tmp for local build artifacts.  Keeping that as a place to clean between runs is reasonable, especially since we don't want to deal with stale build artifacts.

You'll find that with this patch applied that rsync behaves quite swiftly and does exactly the behavior you're saying you want - it only grabs the new things.  It does, however, require the ping for the handshake of what the current directory manifest is.

mahesh-arrcus (approved):

PR #428, opened 2026-01-24
Issue 427: Permit idnits to cache its files between runs

### LINKED ISSUE #427: Permit idnits to cache its files between runs
idnits is hardcoded to stick its working files in $HOME/.idnits.  Since the home directory is ephemeral in the container, download caching is thwarted.

Fix: Create /app/.idnits and symlink that during the clean phase of the build.

jhaas-pfrc:
Fixes #427

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ state/*
 README.ncs
 ncs.conf
 setup.mk
+.idnits
 *~
 #*#
 .#*
--- a/draft/Makefile
+++ b/draft/Makefile
@@ -36,6 +36,10 @@ idnits: $(next).txt
 	$(idnits) $<
 
 clean:
+	# Permit idnits to cache in the mounted directory
+	-mkdir /app/.idnits
+	-ln -s /app/.idnits $(HOME)/.idnits
+
 	-rm -rf ../bin
 	-rm -f $(draft).txt $(draft).html index.html
 	-rm -f $(draft)-[0-9][0-9].*


suehares (commented):
Does this reduce the run time for the yang idnits check?  Seems like a good idea.  Am I missing a downside?

jhaas-pfrc:
It reduces the run time significantly when doing a lot of builds.  I had the misfortune to read how the code was doing this step and think it probably needs a significant update.  However, I expect there's a lot of gotchas on legacy documents and wouldn't propose it to the ietf tools team without discussion.

jhaas-pfrc:
The potential downside is that it clutters the filesystem.  Mahesh might have opinion about the filesystem clutter.  Perhaps centralizing such cache state in a common directory for that purpose makes sense.  This just proves that it's helpful.

suehares:
I'm fine with whatever you or Mahesh decide.   Faster builds = goodness.

mjethanandani:
BTW, I have moved to idnits3 with the latest Docker image for building the draft, which should get rid of the caching needed/used by idnits2.

mjethanandani:
In other words, if you are using build-yang:1.11, kill the instance and remove the image for it, and Docker will pull down build-yang:1.12.

jhaas-pfrc:
I've verified idnits 3 is present in the new image.  It doesn't have the problem I'm complaining about, so closing this PR.  We'll deal with the fallout from the new version in other issues.

PR #429, opened 2026-01-24
Issue 368: Add reference for remove-private-as

### LINKED ISSUE #368: Add citation for remove-private
https://datatracker.ietf.org/doc/html/rfc4116, Section 3.1.2.2 provides a stable reference for remove-private.

jhaas-pfrc:
Fixes: #368

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -263,8 +263,8 @@ submodule ietf-bgp-common {
         "When this leaf is specified, remove private AS numbers from
          updates sent to peers.";
       reference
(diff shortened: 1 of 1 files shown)


suehares (changes requested):
RFC6996 denotes the removal of private 4-octet ASes.   I think you should include this RFC996 for the 4-byte case. 

By the way, does RFC6996 update RFC2270?  Like to chat about the combination.

jhaas-pfrc:
@suehares  RFC 6996 is a much better reference. Changed to use that.

suehares (approved):
RFC6996 works for me.

Issue #432, opened 2026-01-24
Use documentation addresses in examples


jhaas-pfrc:
idnits flagged that there were IP addresses that weren't from the example ranges present in the draft.

jhaas-pfrc:
Reviewing this further, while there seems to be some incorrect addresses which will be fixed by a PR, idnits fails to find the ipv6 example because it's in the appendix.  Similarly, some of the flagged items are actually section numbers that look like a dotted quad.

PR #433, opened 2026-01-24
Issue 432: Fix non-example addresses

jhaas-pfrc:
Partially addresses #432.


suehares (approved):
Good catch on the IPv4 addresses - not being from the right address range. 

Sue

PR #434, opened 2026-01-24
Issue 389: Remove enabled leaf from confederation

### LINKED ISSUE #389: confederation/enabled is redundant?
The confederation configuration includes an "enabled" option.  Typically implementations handle whether the local device is, or is not, part of the confederation via the confederation members.

Note that the member-as description cites "remote" which becomes incorrect if this is updated.

Note that OC doesn't include the enabled leaf.

>         container confederation {
>           description
>             "Configuration options specifying parameters when the
>              local router is within an autonomous system which is
>              part of a BGP confederation.";
>           leaf enabled {
>             type boolean;
>             description
>               "When this leaf is set to true it indicates that
>                the local-AS is part of a BGP confederation.";
>           }
>           leaf identifier {
>             type inet:as-number;
>             description
>               "Confederation identifier for the autonomous system.";
>           }
>           leaf-list member-as {
>             type inet:as-number;
>             description
>               "Remote autonomous systems that are to be treated
>                as part of the local confederation.";
>           }
>         }

jhaas-pfrc:
Fixes: #389

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -322,22 +322,26 @@ module ietf-bgp {
             "Configuration options specifying parameters when the
              local router is within an autonomous system which is
              part of a BGP confederation.";
-          leaf enabled {
-            type boolean;
-            description
-              "When this leaf is set to true it indicates that
-               the local-AS is part of a BGP confederation.";
-          }
+          reference
+            "RFC 5065: Autonomous System Confederations for BGP.";
           leaf identifier {
             type inet:as-number;
             description
-              "Confederation identifier for the autonomous system.";
+              "Confederation identifier for the autonomous system.
+               Setting the identifier indicates that the local-AS is
+               part of the BGP confederation.";
+            reference
+              "RFC 5065: Autonomous System Confederations for BGP,
+               Section 1.2.";
           }
           leaf-list member-as {
             type inet:as-number;
             description
-              "Remote autonomous systems that are to be treated
+              "Autonomous system numbers that are to be treated
                as part of the local confederation.";
+            reference
+              "RFC 5065: Autonomous System Confederations for BGP,
+               Section 1.2.";
           }
         }
         container graceful-restart {
--- a/draft/draft-ietf-idr-bgp-model.xml
+++ b/draft/draft-ietf-idr-bgp-model.xml
@@ -518,8 +518,11 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
       attribute, it will cause the preference for routes, e.g. external vs
       internal to change.</li>
 
-      <li> The attribute 'enabled' in the 'confederation' container. This
-      attribute defines whether a local-AS is part of a BGP confederation.</li>
+      <li> In the 'confederation' container, 'identifier' and 'member-as'
+      defines whether ASes are part of the BGP confederation.  Improperly
+      configuring these attributes may result in BGP session resets by
+      advertising confederation routes to non-confederation peers or
+      non-confederation routes to peers expecting such routes.</li>
 
       <li> Finally, there are a whole set of route selection options such as
       'always-compare-med', 'ignore-as-path-length' that affect the way the


src/yang/ietf-bgp.yang
```diff
@@ -322,22 +322,26 @@ module ietf-bgp {
...
+               part of the BGP confederation.";
+            reference
+              "RFC 5065: Autonomous System Confederations for BGP,
+               Section 1.2.";
           }
           leaf-list member-as {
             type inet:as-number;
             description
-              "Remote autonomous systems that are to be treated
+              "Autonomous system numbers that are to be treated
```
mjethanandani, line 340: The text in the draft says " 'identifier' **and** 'member-as' defines whether ASes are part of the BGP confederation". The description statement for 'identifier' says "Setting the identifier indicates that the local-AS is part of the BGP confederation." Which statement is correct?

suehares (approved):
Good change.  Cleans up the logic for local-AS and AS confederation.

suehares (approved):
Good change cleans up the logic for AS Confederation.

suehares (approved):
Good warning label to go with change to remove Local-AS and put in the identifer and member-as.

src/yang/ietf-bgp.yang
jhaas-pfrc: The distinction is the ASes in the member AS list are part of the confederation that has the identifier in question.

The OC variant that we're cribbing here is trying to say "by setting the identifier, 'local-as' (configured or derived) is part of that member list.

In the JUNOS configuration, it's necessary to define the local-as in the list of the member ASes.  However, automagically pulling it in via this intent is also a valid way to model it.

Where things are one step less pleasant is that it's operationally the case that the member-as list for the confederation MUST be consistently provisioned across all participating routers in the confederation for the feature to work properly.  Misconfiguration of such a consistent list is one of the major points of fragility for this feature.

Ideally, the confed member list is consistent for the reason above.  However, the OC pattern under discussion effectively includes local-as in that list operationally by magic.

@suehares what do you think about this?  Care to help do a survey on implementation configuruation for confederations and help steer how this settles?

suehares: I can do a survey on the implementation configuration for confederations.   I think the Confederation Identifier is an AS, and the member IDs are also ASes.   ETA (Tuesday)

suehares: I've uploaded the AS confederation configurations from Cisco, Huawei, Nokia, and FRR. 

Answering Mahesh's question. 
> The text in the draft says " 'identifier' and 'member-as' defines whether ASes are part of the BGP confederation". The description statement for 'identifier' says "Setting the identifier indicates that the local-AS is part of the BGP confederation." Which statement is correct?

**Answer (from Sue):** Both do.  It is a belt-and-suspenders type of configuration.  Consider the issue of two confederations sharing AS-1.   

Router Bgp AS-1
Confederation 200 AS-1, AS-2, AS-3  # could default to AS-1, so members would be AS-2, AS-3   
Confederation 100 AS-1, AS-4, AS-5  #  could default to AS-1 so members would be AS-4, and AS-

Identifiers identify the confederations.  
Members give a list of EBGP peers that are members of the confederation.  By configuring this nodes, it is clear that AS-1 has a policy that might transfer things from external AS 200 to AS 100.

**Summary of configurations:** 
Setting the identifier in the router bgp context (cisco, huawei, Nokia, FRR) indicates that the bgp peer being configured is part of a confederation denoted by "identifier".  The other AS configured in the Confederation are given by: 
**Cisco:**  
 bgp confederation peers <as-1> <as-2> 

**Huawei:** 
[*Huawei-bgp] confederation peer-as 65002
[*Huawei-bgp] peer 1.1.1.1 as-number 65002 # EBGP to sub-AS 65002

**Nokia:**
config>router# autonomous-system 100
config>router>bgp# confederation 1000 members 100 200 300

**FRR:** 
bgp confederation peers 64513

suehares:
Cisco, Huawei, Nokia and FRR use the concept of 
**confederation identifier** _number_ 
**bgp confederation peers**  _peer1as peer2as_

I'm attaching the exact commands the next comments

suehares:
**Basic Configuration Steps (Cisco IOS):**

**_Define the Confederation Identifier (Public AS):_**
**router bgp** <SUB-AS>
**bgp confederation identifier** <PUBLIC-AS>

**Example:** If the real AS is 100,  but you are using sub-AS 64512,  you use bgp confederation identifier 100.

**_Define Sub-AS Peers:_**
On routers connecting to other sub-ASs within the same confederation, list the other member sub-ASs:
**router bgp** <SUB-AS>
**bgp confederation peers <**SUB-AS2> <SUB-AS3> ...

This tells the router to treat these peers as eBGP (ignoring next-hop-self/iBGP full-mesh rules), but keep the AS_PATH hidden from the outside.

**Configure Neighbors:**
Within the same Sub-AS: Use standard iBGP commands.

**Between Sub-ASs:** 
Use standard eBGP commands (neighbor <ip> remote-as <OTHER-SUB-AS>).

External to Confederation:  Use standard eBGP commands,  the neighbor sees only the identifier.

suehares:
**Huawei Confederation Commands**

huawei sub-as 1 (65001), Sub-as 2 (65002) 
[~Huawei] bgp 65001
[*Huawei-bgp] confederation id 200
[*Huawei-bgp] confederation peer-as 65002
[*Huawei-bgp] peer 1.1.1.1 as-number 65002  # EBGP to sub-AS 65002
[*Huawei-bgp] peer 2.2.2.2 as-number 65001  # IBGP inside 65001
[*Huawei-bgp] commit

[~Huawei] bgp 65002
[*Huawei-bgp] confederation id 200
[*Huawei-bgp] confederation peer-as 65001
[*Huawei-bgp] peer 3.3.3.3 as-number 65001  # EBGP to sub-AS 65001
[*Huawei-bgp] peer 4.4.4.4 as-number 65002  # IBGP inside 65002
[*Huawei-bgp] commit

**Confederation ID:** The official public AS number (200 in this example).
**Confederation Peer-AS:** Lists other sub-ASs within the same confederation.
**Sub-AS Numbers:** Private AS numbers (usually in the 64512–65535 range) used internally.

suehares:
### Key Aspects of Nokia BGP Confederation
**_Configuration & Structure:_** 
The config>router>bgp# context is used to define the confederation ID 
(the main public AS number) and its member ASs.

**Member AS Capabilities:** Up to 15 member ASs can be configured within a single confederation.

Peering Types:
Confederation-IBGP: Internal sessions within a sub-AS.
Confederation-EBGP: Sessions between different sub-ASs inside the same 
confederation. These function like EBGP but do not require strict IGP synchronization.

AS Path Handling: When advertising routes outside, 
the confederation ID is prepended, while inner sub-AS numbers 
(often private ASNs) are stripped.

**Example Configuration:**
config>router# autonomous-system 100
config>router>bgp# confederation 1000 members 100 200 300
(Here 1000 is the public AS, 100/200/300 are private sub-ASs). 

**Benefits**
Scalability: Eliminates the need for a full mesh of IBGP sessions, improving network performance.

suehares:
### Free Range Routing 

**1. Define main AS**
router bgp <Main_Public_ASN>
bgp confederation identifier <Main_Public_ASN>
router bgp <Sub_ASN>

**2. Define Sub-ASs: Assign each router to a private sub-AS (e.g., 64512-65534).**
text
router bgp <Sub_ASN>

**3. Define Confederation Peers:** 
List all other sub-ASs within the confederation on all routers.
text
router bgp <Sub_ASN>
bgp confederation peers <Sub_ASN1> <Sub_ASN2> 

**4. Configure BGP Neighbors:**
Internal (iBGP): Peers within the same sub-AS use standard iBGP config.
External (Confed-EBGP): Peers in different sub-ASs are treated as EBGP (often using ebgp-multihop if using loopbacks). 

**Example Configuration**
Imagine a main AS 100, split into sub-AS 64512 and 64513.

**Router A (Sub-AS 64512):**
text
router bgp 64512
 bgp confederation identifier 100
 bgp confederation peers 64513
 neighbor 10.0.0.2 remote-as 64512 ! iBGP peer
 neighbor 192.168.0.2 remote-as 64513 ! Confed-EBGP peer
 
**Router B (Sub-AS 64513):**
text
router bgp 64513
 bgp confederation identifier 100
 bgp confederation peers 64512
 neighbor 192.168.0.1 remote-as 64512 ! Confed-EBGP peer
 
 Note: 
 Confederation eBGP sessions preserve next-hop, MED, and Local Preference information, acting like iBGP.

jhaas-pfrc:
Given what we have for the various supplied configs, let's stick with what we have.  Thanks for digging out the examples for the audit trail.

PR #435, opened 2026-01-26
Issue 381: Update 'instance' security considerations.

### LINKED ISSUE #381: Security considerations for identifier may need update
>  The attribute 'identifier'. If a user is allowed to change this attribute, it will have the net effect of this routing instance re-advertising all its routes

The identifier is exchanged in the BGP open message.  The only way for the remote peer to pick it up is a session bounce.  Bouncing the session is what Junos does.

The question is whether this bounce behavior is universal enough to update.  This begs the question of whether the current text was intentional or simply confused.

jhaas-pfrc:
Also, make some updates to 'as' security considerations.

Fixes: #381


suehares (approved):
Better language for the security warnings.   Thank you for fixing this language.

PR #437, opened 2026-01-26
Issue 382: Reorg and update text for secure-session

### LINKED ISSUE #382: Security considerations for authentication
> The list of neighbors, and their attributes. Allowing a user to read these attributes, in particular the address/port information may allow a malicious user to launch an attack at the particular address/port

One would think mentioning authentication as being important to protect should be done...

jhaas-pfrc:
Fixes: #382


mjethanandani (approved):

suehares (approved):
Nice revision.  I'm going to steal the ideas for SDWAN.

PR #438, opened 2026-01-26
Issue 397: Remove negotiated-hold-time

### LINKED ISSUE #397: Reconsider op state for hold-time leaf
The text for read-only get on the hold-time is effectively retriving the leaf negotiated hold-time. 

```
    leaf hold-time {
      type uint16 {
        range "0 | 3..65535";
      }
      units "seconds";
      default "90";
      description
        "Time interval (in seconds) for the HoldTimer established
         with the peer.  When read as operational data (ro), the
         value of this object is calculated by this BGP speaker,
         using the smaller of the values in hold-time that was
         configured (rw) in the running datastore and the Hold Time
         received in the OPEN message.

         This value must be at least three seconds
         if it is not zero (0).

         If the Hold Timer has not been established
         with the peer this object MUST have a value
         of zero (0).

         If the configured value of hold-time object was
         a value of (0), then when read this object MUST have a
         value of (0) also.";
      reference
        "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 4.2,
         RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 10.";
    }
    leaf negotiated-hold-time {
      type uint16;
      config false;
      description
        "The negotiated hold-time for the BGP session";
      reference
        "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 4.2,
         RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 10.";
    }
```

jhaas-pfrc:
As noted in the issue, since IETF access protocols can distinguish between operational state vs. configuration state, we don't need the op-only "negotiated-hold-time".

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -143,15 +143,6 @@ submodule ietf-bgp-common {
         "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 4.2,
          RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 10.";
     }
-    leaf negotiated-hold-time {
-      type uint16;
-      config false;
-      description
-        "The negotiated hold-time for the BGP session";
-      reference
-        "RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 4.2,
-         RFC 4271: A Border Gateway Protocol 4 (BGP-4), Section 10.";
-    }
     leaf keepalive {
       type uint16 {
         range "0..21845";


mjethanandani:
> As noted in the issue, since IETF access protocols can distinguish between operational state vs. configuration state, we don't need the op-only "negotiated-hold-time".

That is correct. The 'hold-timer' node will be represented in the <running> datastore as the configured (rw) value, and in the <operational> datastore as a "negotiated hold timer" (ro).

mjethanandani (approved):

suehares (commented):
Negotiated-hold-time - should be operational-state, not configured state. 

Ack for removal.

PR #439, opened 2026-01-26
Issue 390: Split errors groupings into new submodule

### LINKED ISSUE #390: Consider moving the bgp-errors / common-errors groupings to the common submodule
The goal is to keep grouping clutter out of the general structure of the module.

jhaas-pfrc:
Fixes #390

### DIFF
--- a/src/yang/ietf-bgp-common-errors.yang
+++ b/src/yang/ietf-bgp-common-errors.yang
@@ -1,0 +1,181 @@
+submodule ietf-bgp-common-errors {
+  yang-version 1.1;
+  belongs-to ietf-bgp {
+    prefix bgp;
+  }
+
+  import iana-bgp-notification {
+    prefix bn;
+    reference
+      "RFC XXXX: YANG Model for Border Gateway Protocol (BGP-4).";
+  }
+  import ietf-yang-types {
+    prefix yang;
+    reference
+      "RFC 6991: Common YANG Data Types.";
+  }
+
+  organization
+    "IETF IDR Working Group";
+  contact
+    "WG Web:   <http://datatracker.ietf.org/wg/idr>
+     WG List:  <idr@ietf.org>
+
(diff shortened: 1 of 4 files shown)


jhaas-pfrc:
We may need to make submodules go away.  If so, perhaps this move is the wrong thing?  I'm not sure what our plan is for groupings like this as part of cleanup.

src/yang/ietf-bgp-common-errors.yang
```diff
@@ -0,0 +1,181 @@
...
+
+  import iana-bgp-notification {
+    prefix bn;
+    reference
+      "RFC XXXX: YANG Model for Border Gateway Protocol (BGP-4).";
+  }
+  import ietf-yang-types {
+    prefix yang;
+    reference
+      "RFC 6991: Common YANG Data Types.";
```
mjethanandani: This is now RFC 9911.

src/yang/ietf-bgp-common-errors.yang
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+
+  grouping bgp-encapsulated-errors-common {
+    description
+      "BGP NOTIFICATION state that is common in the sent and received
+       direction that has been encapsulated in a CEASE/HARD RESET
+       NOTIFICATION.
+
+       Note that these leaves are only present when carrying the RFC
```
mjethanandani: More of a nit, as YANG uses leafs as a plural for leaf.

s/leaves/leafs/

mjethanandani:
> We may need to make submodules go away. If so, perhaps this move is the wrong thing? I'm not sure what our plan is for groupings like this as part of cleanup.

The plan is to continue the use of submodules.

suehares (approved):
Approve with Mahesh's change to RFC9911 and leafs (although my English language spider senses tingled).

PR #440, opened 2026-01-26
Issue 403: Update as-path-prepend

### LINKED ISSUE #403: as-path-prepend inconsistencies
The YANG we currently have includes a repeat-n for the common use case for prepending.
However, it also includes an asn to prepend.  Okay, still somewhat common in implementations.
Messily, it includes a leaf-list of asns to prepend.  Okay, still somewhat common.

Describing how the leaf-list use case and the repeat-n interacts needs work.

```
      container set-as-path-prepend {
        description
          "Action to prepend local AS number to the AS-path a
           specified number of times";

        leaf repeat-n {
          type uint8 {
            range "1..max";
          }
          description
            "Number of times to prepend the AS number to the AS
             path. The value should be between 1 and the maximum
             supported by the implementation.";
        }
        leaf-list asn {
          type inet:as-number;
          description
            "AS number to prepend to the AS path.";
        }
      }
```

jhaas-pfrc:
Fixes: #403 

Note: The inspiration and some of the description text was copied from the openconfig equivalent nodes.

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -661,16 +661,17 @@ module ietf-bgp-policy {
           }
           description
             "Number of times to prepend the AS number to the AS
-             path. If no value is specified by the asn leaf, the local
-             AS number is used. The value should be between 1 and the
-             maximum supported by the implementation.";
+             path. If no value is specified by the asn or use-last-as
+             leafs, the local AS number is used. The value should be
+             between 1 and the maximum supported by the
+             implementation.";
         }
         leaf asn {
           type inet:as-number;
           must "../repeat-n and not(../use-last-as)";
           description
             "The AS number to prepend to the AS path. If neither this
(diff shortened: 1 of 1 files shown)


suehares (changes requested):
"Number of times to prepend the AS number to the AS
             path. **If no value is specified by the asn leaf, the local
             AS number is used. ** The value should be between 1 and the
             maximum supported by the implementation.";

The text is wrong in the 2nd sentence.  Value is the number of times the AS is prepended. 
I think you need to rewrite this section. 

              "Number of times to prepend the AS number to the AS
             path.  The value should be between 1 and the
             maximum supported by the implementation.";

 I think this is what you wanted here.    I think you need to rework this commit.

jhaas-pfrc:
@suehares I believe I have addressed your review comments.  Please re-review.

suehares (approved):
Works for me.

suehares (approved):
text Works for me.

PR #454, opened 2026-02-02
Issue 420: Add treat-as-withdraw to global scope

### LINKED ISSUE #420: treat-as-withdraw is missing global scope
The leaf treat-as-withdraw can't be set at global scope.
In general, this is a reminder to audit per group/neighbor config state for equivalence at global.

jhaas-pfrc:
Fixes: #420

Additionally, move the location of this to be next to enforce-first-as.


suehares (approved):
Treat-as-withdraw  - should be included as basic BGP.

PR #455, opened 2026-02-02
Issue 453: Remove defaults for holdtime

### LINKED ISSUE #453: Consider removing default from hold-time
We're coming to see that specifying "default" in YANG nodes, even when protocol recommended, is problematic for implementations. By having a default node, the implementation's own default must either be created through a deviation or the operator is forced to override it each place they want to use it.

jhaas-pfrc:
Fixes: #453


suehares (approved):
This moves the yang module to no defaults.   I am fine with this for configuration.

Issue #462, opened 2026-02-02
OpenConfig parity - policy comparators and value


jhaas-pfrc:
OpenConfig's policy has a generic "value" element to be compared when comparing vs. counters.  The IETF model has leafs that are named largely as a duplicate inside of their policy container.

Evaluate whether the OpenConfig model is simpler.

jhaas-pfrc:
Answering the question from the inside of our own code, things like "local-pref" and "med" use value.

The right answer here appears to be to clean up and use "value".

PR #463, opened 2026-02-02
Issue 461: Add extended-community-count policy

### LINKED ISSUE #461: OpenConfig parity: extended-community-count policy

jhaas-pfrc:
Addresses OpenConfig parity. Fixes: #461

Note that the content of this container may be impacted by the resolution of issue #462.

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -463,6 +463,21 @@ module ietf-bgp-policy {
         uses equality-operator;
       }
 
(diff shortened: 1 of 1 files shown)


suehares (approved):
Nice addition.

keyur-arrcus:
I see the addition of value. Do we need to worry about extended-communities being different but of same count?

jhaas-pfrc:
> I see the addition of value. Do we need to worry about extended-communities being different but of same count?

They are different leafs and easy to distinguish.

PR #464, opened 2026-02-02
Issue 458: Move advertise-inactive-paths

### LINKED ISSUE #458: advertise-inactive-routes probably doesn't belong in route-selection-options

jhaas-pfrc:


suehares (commented):
So you are moving advertise-inactive-routes from 
grouping neighbor-group-config  to
grouping route-selection-options 

Can you explain moire?

jhaas-pfrc:
Here is the location before the patch:

```
    +--rw bgp
       +--rw global!
       | 
       |  +--rw route-selection-options
       |  |  |  +--rw advertise-inactive-routes?    boolean
       | 
       |  +--rw afi-safis
       |  |  +--rw afi-safi* [name]
       |  |     +--rw route-selection-options
       |  |     |  +--rw always-compare-med?           boolean
       |  |     |  +--rw ignore-as-path-length?        boolean
       |  |     |  +--rw external-compare-router-id?   boolean
       |  |     |  +--rw advertise-inactive-routes?    boolean
```

Here's the after:
```
       +--rw peer-groups
       |  +--rw peer-group* [name]

       |  |  +--rw advertise-inactive-routes?   boolean
       |  |  +--rw add-paths {bt:add-paths}?
[...]
       +--rw neighbors
       |  +--rw neighbor* [remote-address]
       |     +--rw advertise-inactive-routes?   boolean
       |     +--rw add-paths {bt:add-paths}?

```

suehares:
### Summary 
Cisco, Huawei - seem to me to be at the top level, 
Juniper - at the group/peer level 
Nokia - at the group/peer level with policy involved 

If this is OC location, why did they move it here?  Just asking.  Equal pro/con if you just look  at config. 

### Cisco 

Cisco commands that seem to replicate Jeff's placement. 

router bgp <AS-NUMBER>
 bgp suppress-inactive

**_makes sure that the command_** 
redistribute <protocol> [process-id] [route-map <map-name>]
or 
bgp redistribute-internal

**_does not redistribute_** 

Seems to be at the 

### JunOS
This stsatement allows the advertisement of inactive (opposite cisco suppress inactive) 

set protocols bgp group <group-name> advertise-inactive

This is active at the BGP group level or Peer level.  It does not seem. 

### Huawei 
[~Huawei-bgp] ipv4-family unicast
[~Huawei-bgp-af-ipv4] import-route ospf 1
[~Huawei-bgp-af-ipv4] import-route direct allow-ibgp

To redistribute inactive BGP routes (routes not selected as best) on Huawei devices, you must specifically configure the import-route command to include them, as BGP typically only distributes active, best-path routes

This set of commands seems to be at BGP level.  

### Nokia 
configure router policy-options
    begin
    prefix-list "inactive-routes"
        prefix 10.0.0.0/24 exact
    exit
    policy-statement "redistribute-inactive"
        entry 10
            from
                prefix-list "inactive-routes"
                protocol static  # Example: include static routes
            exit
            action accept
            exit
        exit
    commit

configure router bgp
    export "redistribute-inactive"
    # Optional: For specific cases, use:
    # advertise-inactive

mjethanandani:
And since Arrcus uses OC models, the advertise-inactive-routes falls under global.

jhaas-pfrc:
> And since Arrcus uses OC models, the advertise-inactive-routes falls under global.

OC's currently location is part of the route selection options in this grouping:
bgp-common-route-selection-options-config

src/yang/ietf-bgp-common.yang
```diff
@@ -409,6 +409,18 @@ submodule ietf-bgp-common {
...
+
+    leaf advertise-inactive-routes {
+      type boolean;
+      default "false";
+      description
+        "Advertise inactive routes to external peers.  The default
+         is to only advertise active routes.";
+      reference
+        "I-D.ietf-idr-best-external: Advertisement of the best
+         external route in BGP.";
```
keyur-arrcus, line 421: This looks good.  That has to be per neighbor per afi/safi based.

keyur-arrcus (approved):
Looks good.

keyur-arrcus:
One thing we could possibly do is keep it at a afi/safi level and then have a neighbor based knob to selectively announce it out.

PR #465, opened 2026-02-02
Issue 394: Change in-updates-elapsed-time

### LINKED ISSUE #394: Reconsider in-update-elapsed-time
The in-update-elapsed-time is the "how long from now" vs. the last received time for an update.  This means, by definition, it's constantly updating.  This isn't good for streaming telemetry.

```
              leaf in-update-elapsed-time {
                type yang:gauge32;
                units "seconds";
                description
                  "Elapsed time (in seconds) since the last BGP
                   UPDATE message was received from the peer.
                   Each time in-updates is incremented,
                   the value of this object is set to zero (0).";
                reference
                  "RFC 4271: A Border Gateway Protocol 4 (BGP-4),
                   Section 4.3
                   RFC 4271: A Border Gateway Protocol 4 (BGP-4),
                   Established state.";
              }
```

jhaas-pfrc:
Fixes: #394

An elapsed time will always be continuously changing from the perspective of on-change telemetry.  Thus, it needs to be a fixed timestamp of last update.

Change type to be yang:timeticks and name to reflect that it's no longer elapsed: in-updates-time.

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -639,14 +639,14 @@ module ietf-bgp {
                   "RFC 7606: Revised Error Handling for BGP UPDATE
                    Messages, Section 2.";
               }
-              leaf in-update-elapsed-time {
-                type yang:gauge32;
-                units "seconds";
+              leaf in-update-time {
+                type yang:timeticks;
                 description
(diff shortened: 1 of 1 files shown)


suehares (approved):
Change looks good to me.

PR #467, opened 2026-02-02
Issue 460: Update prefix statistics description and leafs

### LINKED ISSUE #460: OpenConfig parity - prefix counters
The current model covers "prefixes" for
- received
- sent
- installed

The OpenConfig model also covers received-pre-policy.

Compare the two models on this point and see if a feature should be added.

jhaas-pfrc:
Fixes: #460 

Audit vs. the OpenConfig model suggested a few things:
- It would be useful to have a pre-policy received leaf.
- The descriptive text could use clarification
- The idea of "installed" has drifted between the IETF and the OpenConfig implementation.

OpenConfig conflates installed routes (the routing table) with Loc-RIB. A consequence of this is that their statistics don't cover the best-path counts; rather it uses "number of routes installed for forwarding".

The "paths used for forwarding" is still a useful metric when ECMP or other load balancing is implemented.  This PR splits the two cases of loc-rib contribution and multi-path contribution.

### DIFF
--- a/src/yang/ietf-bgp-neighbor.yang
+++ b/src/yang/ietf-bgp-neighbor.yang
@@ -83,27 +83,60 @@ submodule ietf-bgp-neighbor {
            a session restart may be required in order to negotiate
            the new capability.";
       }
+
       container prefixes {
         config false;
         description
           "Prefix counters for the AFI/SAFI in this BGP session";
+
+        leaf received-pre-policy {
+          type yang:zero-based-counter32;
+          description
+            "The number of prefixes that are received from the
+             neighbor before applying any policies. This count is the
+             number of prefixes present in the pre-policy
+             adj-rib-in-pre table for this AFI/SAFI.";
+        }
+
         leaf received {
           type yang:zero-based-counter32;
           description
-            "The number of prefixes received from the neighbor";
+            "The number of prefixes that are received from the
+             neighbor after applying any policies. This count is the
+             number of prefixes present in the adj-rib-in-post table
+             for this AFI/SAFI.";
         }
+
         leaf sent {
           type yang:zero-based-counter32;
           description
-            "The number of prefixes advertised to the neighbor";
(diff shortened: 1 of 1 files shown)


suehares (approved):
Very good change to split loc-rib and multi-path contribution. 
It is important to get the following separated: 
1) before policy prefixes count of prefixes received, 
2) after policy runs - count of prefixes received, 
3) number of prefixes sent to neighbor after running policy 
4) number of prefixes for AFI/SAFI installed in loc-rib. 

Very good!

jhaas-pfrc:
Thanks for the approval, Sue.  This one I'd also like @keyur-arrcus 's eyes on.

mjethanandani (approved):

src/yang/ietf-bgp-neighbor.yang
```diff
@@ -83,27 +83,60 @@ submodule ietf-bgp-neighbor {
...
+
+         leaf installed-multiple-paths {
+           type yang:gauge32;
+           description
+             "The number of prefixes received from the neighbor for
+              this AFI/SAFI that are used for forwarding.  These
+              include route that are:
+
+              - 'elibigle-route' is true.
+              - 'best-path' true, or the implementation installs this
```
keyur-arrcus, line 134: Should we make sure that the gauge is incremented once if the path is best & multipath?

keyur-arrcus (commented):
Looks good except for the best path and multipath question I posted for the gauge.

Keyur and Mahesh!

src/yang/ietf-bgp-neighbor.yang
jhaas-pfrc: The 'best-path' is true *OR* ... is intended to capture this.  Did you have different verbiage in mind?

jhaas-pfrc: @keyur-arrcus explicit ack here, thanks.

PR #470, opened 2026-02-08
Issue 462: Update policy to be more consistent in naming

### LINKED ISSUE #462: OpenConfig parity - policy comparators and value

jhaas-pfrc:
Fixes: #462

When the pattern of policy is of the form
container foo {
leaf foo; // To be matched
uses compare-oper;
}

the second instance of foo is cognitive noise.

As we have elswhere, simply label the leaf "value".

### DIFF
--- a/src/yang/ietf-bgp-policy.yang
+++ b/src/yang/ietf-bgp-policy.yang
@@ -455,7 +455,7 @@ module ietf-bgp-policy {
           "Value and comparison operations for conditions based on
            the number of communities in the route update.";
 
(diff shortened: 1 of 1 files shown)


suehares (approved):
fine with the change if matches normal yang flow.  I do find the normal yang "value" method counter-intutive.

PR #475, opened 2026-02-09
Issue 400: Remove default for treat-as-withdraw

### LINKED ISSUE #400: Reconsider treat-as-withdraw default
When false, this effectively says "rfc 7606 isn't on by default".

```
    leaf treat-as-withdraw {
      type boolean;
      default "false";
      description
        "Specify whether erroneous UPDATE messages for which the NLRI
         can be extracted are treated as though the NLRI is withdrawn
         - avoiding session reset";
      reference
        "RFC 7606: Revised Error Handling for BGP UPDATE Messages.";
    }
```

jhaas-pfrc:
The treat-as-withdraw leaf had a default for false.  Defaults manifest in the model as nodes when you don't explicitly set them.  These days we want treat-as-default to be on by default, when supported.  However, we also have been moving to a pattern where we don't add defaults to the model to permit implementations to have their own internal defaults.

Fixes: #400

### DIFF
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -251,11 +251,11 @@ submodule ietf-bgp-common {
     }
     leaf treat-as-withdraw {
       type boolean;
-      default "false";
       description
         "Specify whether erroneous UPDATE messages for which the NLRI
          can be extracted are treated as though the NLRI is withdrawn
-         - avoiding session reset";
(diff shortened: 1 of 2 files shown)


suehares (commented):
Commit looks good. Need feedback on Keyur's implementation experience as well as other folks.

jhaas-pfrc:
@keyur-arrcus note Sue is asking for explicit review.

PR #476, opened 2026-02-09
Issue #457: Remove ietf-bgp-common-structure sub-module

### LINKED ISSUE #457: Reconsider module name or org for "structure"
The IETF YANG module had some of the early work from OpenConfig and includes as an artifact the name of the "structure" submodule.

In OpenConfig, structure models are needed because they split configuration and operational state in different subtrees with common groupings in most cases between them.  However, since IETF doens't do this split, there's no need for that split.

The "structure" in OpenConfig models effectively are models that have no leaf nodes, only containers for groupings that will get them.  

Consider either renaming the existing structure model or moving its contents to the common model.

jhaas-pfrc:
Remove structure module and spread its comments elsewhere in the existing common module.

Fixes: #457


jhaas-pfrc:
@mjethanandani please review vs. best practices for where groupings go vs. their users.

suehares (approved):
You have doe the split correctly.   Why was this necessary?  Not a block, just a quesiton.

mjethanandani (commented):
Is there a reason to move the "structure" groupings into ietf-bgp-common and not ietf-bgp-neighbor? The name of the groupings seems to suggest them as "neighbor" groupings.

jhaas-pfrc:
> You have doe the split correctly. Why was this necessary? Not a block, just a quesiton.

The "structure" model came in 2016 in -02 of the current draft.  I have no idea what the motivation was at the time.

jhaas-pfrc:
> Is there a reason to move the "structure" groupings into ietf-bgp-common and not ietf-bgp-neighbor? The name of the groupings seems to suggest them as "neighbor" groupings.

You'll note that the groupings are moved into the model where the external uses was done.  (See updated uses entries in the diff.)  So, the grouping is now closer to where the definition is.

Please feel free to review the contents of the common grouping to see if further cleanup is useful recursively.  The high level practice we seem to have moved into is we construct groupings for things that are singly included because it helps keep the underlying structure of the main modules clean.

PR #480, opened 2026-04-03
Consider making some of the submodules as modules.

### LINKED ISSUE #479: Consider removing the submodule structure in the YANG module
Submodules in YANG belong to a single parent module. The current BGP model uses a submodule structure to organize the code. While that is a valid structure, it does lead to the issue that groupings defined in the submodule are not usable outside the parent module.

There have been a few requests for the ability to use groupings from the model. To be able to do that, the model will have to be reorganized, remove submodules and make them standalone modules that are imported rather than included by the parent module.

Question to @jhaas-pfrc, @suehares and @keyur-arrcus as chairs of IDR, this is a major change. Does it mean we need another WGLC?

mjethanandani:
Address issue #479

### DIFF
--- a/draft/Makefile
+++ b/draft/Makefile
@@ -75,8 +75,8 @@ rib: ../src/yang/ietf-bgp-rib.yang ../src/yang/ietf-bgp-rib-attributes.yang ../s
 	-mkdir -p ../bin/submodules
 	-ln -s /app/yang-parameters ../bin/yang-parameters 
 	sed -e"s/YYYY-MM-DD/$(shell date +%Y-%m-%d)/" ../src/yang/ietf-bgp-rib.yang > ../bin/submodules/ietf-bgp-rib\@$(shell date +%Y-%m-%d).yang
-	sed -e"s/YYYY-MM-DD/$(shell date +%Y-%m-%d)/" ../src/yang/ietf-bgp-rib-attributes.yang > ../bin/submodules/ietf-bgp-rib-attributes\@$(shell date +%Y-%m-%d).yang
-	sed -e"s/YYYY-MM-DD/$(shell date +%Y-%m-%d)/" ../src/yang/ietf-bgp-rib-tables.yang > ../bin/submodules/ietf-bgp-rib-tables\@$(shell date +%Y-%m-%d).yang
+	sed -e"s/YYYY-MM-DD/$(shell date +%Y-%m-%d)/" ../src/yang/ietf-bgp-rib-attributes.yang > ../bin/ietf-bgp-rib-attributes\@$(shell date +%Y-%m-%d).yang
+	sed -e"s/YYYY-MM-DD/$(shell date +%Y-%m-%d)/" ../src/yang/ietf-bgp-rib-tables.yang > ../bin/ietf-bgp-rib-tables\@$(shell date +%Y-%m-%d).yang
 
 .INTERMEDIATE: $(draft).xml
 %.xml: %.md
--- a/draft/draft-ietf-idr-bgp-model.xml
+++ b/draft/draft-ietf-idr-bgp-model.xml
@@ -654,6 +654,17 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
             prefix: brt
             reference: RFC XXXX
 
+            name: ietf-bgp-rib-attributes
+            namespace:
+              urn:ietf:params:xml:ns:yang:ietf-bgp-rib-attributes
+            prefix: bgp-rib-attr
+            reference: RFC XXXX
+
+            name: ietf-bgp-rib-tables
+            namespace: urn:ietf:params:xml:ns:yang:ietf-bgp-rib-tables
+            prefix: bgp-rib-tbl
+            reference: RFC XXXX
+
             name: ietf-bgp-capabilities
             belongs to: ietf-bgp
             reference: RFC XXXX
@@ -675,14 +686,6 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
             reference: RFC XXXX
 
             name: ietf-bgp-neighbors
-            belongs to: ietf-bgp
-            reference: RFC XXXX
-
-            name: ietf-bgp-rib-attributes
-            belongs to: ietf-bgp
-            reference: RFC XXXX
-
-            name: ietf-bgp-rib-tables
             belongs to: ietf-bgp
             reference: RFC XXXX
 
@@ -727,10 +730,6 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
           neighbor context.</li>
 
           <li>ietf-bgp-rib - grouping for representing BGP RIBs.</li>
-          <li>ietf-bgp-rib-attributes - common data definitions for
-          BGP attributes used in BGP RIB tables.</li>
-          <li>ietf-bgp-rib-tables - structural data definitions for
-          BGP routing tables.</li>
         </ul></t>
 
       <t>Additionally, the above modules include the following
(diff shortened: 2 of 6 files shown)


jhaas-pfrc (changes requested):
This looks fine, but shouldn't we also take our own test module, update that, and validate there?

suehares (approved):
Looks good to me.  Thanks for the work!

suehares:
Jeff - I will review test module for content.  Beyond that I cannot help until I gain more skill in yang validation.

mjethanandani:
> This looks fine, but shouldn't we also take our own test module, update that, and validate there?

If by our own test module you mean ietf-bgp-commands.yang, which BTW, we have not touched for 5 years, still compiles.

jethanandan-a01:ietf-bgp-yang mahesh$ yanglint -p bin/yang-parameters/ -p bin/submodules/ -p bin/  bin/ietf-bgp-commands\@2026-04-10.yang 
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-common) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-common-structure) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-neighbor) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-neighbor) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-neighbor) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-rib) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-rib) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (ietf-bgp-rib-tables) are not necessary.

jethanandan-a01:ietf-bgp-yang mahesh$ pyang --lint --strict --canonical -p bin/yang-parameters -p bin/submodules/ -p bin/ -f tree bin/ietf-bgp-commands\@2026-04-10.yang 
bin/ietf-bgp-commands@2026-04-10.yang:1: error: RFC 8407: 4.8: statement "module" must have a "contact" substatement
bin/ietf-bgp-commands@2026-04-10.yang:1: error: RFC 8407: 4.8: statement "module" must have a "organization" substatement
bin/ietf-bgp-commands@2026-04-10.yang:1: error: RFC 8407: 4.8: statement "module" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:32: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:45: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:58: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:71: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:84: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:96: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:109: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:122: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:135: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
bin/ietf-bgp-commands@2026-04-10.yang:148: error: RFC 8407: 4.14: statement "augment" must have a "description" substatement
module: ietf-bgp-commands

  augment /rt:routing/rt:control-plane-protocols/rt:control-plane-protocol/bgp:bgp/bgp:rib/bgp:afi-safis/bgp:afi-safi/bgp:ipv4-unicast/bgp:loc-rib/bgp:routes/bgp:route:
    +--ro attributes
    |  +--ro origin?                bt:bgp-origin-attr-type
    |  +--ro as-path
    |  |  +--ro segment* []
    |  |     +--ro type?     bt:as-path-segment-type
    |  |     +--ro member*   inet:as-number
    |  +--ro as4-path
    |  |  +--ro segment* []
    |  |     +--ro type?     bt:as-path-segment-type
    |  |     +--ro member*   inet:as-number
    |  +--ro next-hop?              inet:ip-address
    |  +--ro link-local-next-hop?   inet:ipv6-address
    |  +--ro med?                   uint32
    |  +--ro local-pref?            uint32
    |  +--ro aggregator
    |  |  +--ro as?           inet:as-number
    |  |  +--ro identifier?   yang:dotted-quad
    |  +--ro aggregator4
    |  |  +--ro as4?          inet:as-number
    |  |  +--ro identifier?   yang:dotted-quad
    |  +--ro atomic-aggregate?      boolean
    |  +--ro originator-id?         yang:dotted-quad
    |  +--ro cluster-list*          yang:dotted-quad
    |  +--ro aigp-metric?           uint64
    +--ro community*            union
    +--ro ext-community*        bct:bgp-ext-community-type
    +--ro ext-community-raw*    string
    +--ro large-community*      bct:bgp-large-community-type
    +--ro last-modified?        yang:timeticks
    +--ro eligible-route?       boolean
    +--ro ineligible-reason?    identityref
    +--ro unknown-attributes
    |  +--ro unknown-attribute* [attr-type]
    |     +--ro optional?     boolean
    |     +--ro transitive?   boolean

rest of the tree is snipped for brevity reasons.

mjethanandani:
Something else that I realized along the way. The request from YANG Push to make sub-modules into modules so they can use the groupings is redundant. 

It turns out that one can use groupings from sub-modules by importing the parent module and then using the prefix of the parent module to reference the groupings one needs to include.

That is also the reason why our own test module still compiles. Because it uses the prefixes of the parent module.

Therefore, the question before us is, do we still need this change?

jhaas-pfrc:
> Something else that I realized along the way. The request from YANG Push to make sub-modules into modules so they can use the groupings is redundant.
> 
> It turns out that one can use groupings from sub-modules by importing the parent module and then using the prefix of the parent module to reference the groupings one needs to include.
> 
> That is also the reason why our own test module still compiles. Because it uses the prefixes of the parent module.
> 
> Therefore, the question before us is, do we still need this change?

This was exactly one of my motivations to ensure the commands module (yes, I realize it's old - but it's still important to make sure it worked) was in working order.  I think you've proven that it is unnecessary to expose the submodules.  I was unclear whether we'd have name spacing issues for non-bgp work, but this seems to suggest the answer to that is no.

I think the remaining question to answer for the NMOP work is whether there's other negative considerations to force the equivalent of "import all of BGP".  Could this introduce extra "weight" to clients that want to consume the portions of the schema that the NOP work wants that is unneeded?

jhaas-pfrc:
From Leonardo Rodini on IDR 13 April, 2026:
#
Hi Mahesh, Jeff,
 
Thanks for your time spent in investigating this and restructuring the modules.
 
I have tested on my side and yes indeed I was also able to reuse the groupings when including the whole ietf-bgp model. This works, however means that we are forced also to import all the modules including the ones only config related which we would not need. For this reason, to have a cleaner solution with less overhead we still would prefer the change into modules which we then can selectively import.
 
Regarding the PR I see you only targeted rib-attributes and rib-tables and left all others as submodules. Any reason for preferring submodules for the remaining ones? I ask because we would also like to reuse groupings in the ietf-bgp-capabilities submodule (for modeling BMP peer-up messages containing BGP OPENs). From last discussions I got the impression you would attempt a general refactoring, but this was probably a misunderstanding from our side. Anyway, if you decide to go on with the change into modules then we'd need ietf-bgp-capabilities as well.
 
Let us know what you think and have a nice day!

Cheers,
Leonardo  
 
# 
The request thus remains to expose the modules more completely.

suehares (approved):
thank you for making these changes.

PR #485, opened 2026-05-28
Permit neighbor key for BGP peers to be distinct from remote-address

### LINKED ISSUE #484: Permit the neighbor key for BGP peers to be distinct from the remote-address.
On 11 April, 2026, Maria Matejka on IDR suggested a way to address her concern for a more flexible neighbor key:

> We can not deviate the key, at least nobody around Netmod was able to tell me how. We would have to deviate the whole neighbor list, and consequentially probably everything which leaf-refs that. Or, we could have deviated the remote-address, which works but brings other problems with the remote-address suddenly not being a remote-address, actually.
> 
> What may work tho, is defining the neighbor key as a separate item which would by default be only the remote address, and that item could then be deviated / augmented much easier.
> 

```
container neighbors {
  list neighbor {
    key "neighbor-key";
    leaf neighbor-key {
      type leafref {
    path "remote-address";
      }
    }
    leaf remote-address {
      type inet:ip-address;
    }
    ...
  }
  ...
}
```

mjethanandani:


suehares:
This matches the discussion on the list.  Jeff sent a note to Maria (BIRD) for confirmation.  I'm going to OK.  Jeff is going to oK after Maria responds. 

Cheers

suehares (approved):
I'm going to approve because this matches our email discussion.  Jeff sent email to Maria (at Bird) to get her to confirm this change.  He will confirm the change once Maria responds. 

Sue

jhaas-pfrc (approved):
Approved, but please hold on merging until confirmation from BIRD developers.

marenamat:
Looks good to me, thanks for that and thanks Jeff for the heads-up. As far as I know the YANG, this should work for us.

PR #487, opened 2026-06-07
Issue #481 Clean up description for last-established

### LINKED ISSUE #481: Leaf last-established is yang:date-and-time but the description speaks about seconds since epoch.
Reported by Maria Matejka on IDR vs. -19

jhaas-pfrc:
Remove references to unix epoch and UTC.
Closes #481

### DIFF
--- a/src/yang/ietf-bgp.yang
+++ b/src/yang/ietf-bgp.yang
@@ -534,13 +534,11 @@ module ietf-bgp {
             description
               "This timestamp indicates the time that the BGP session
                last transitioned in or out of the Established state.
-               The value is the timestamp in seconds relative to the
-               Unix Epoch (Jan 1, 1970 00:00:00 UTC).
 
                The BGP session uptime can be computed by clients as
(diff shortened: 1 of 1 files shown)


jhaas-pfrc:
My one concern here is that there's an operational likely preference that the TZ output should be 'Z' for easier comparison across a multi-geography deployment.  However, the YANG type doesn't provide such recommendations.

@mjethanandani  is there any such templated language here recommended by the YANG doctors?

suehares (approved):

mjethanandani (approved):

mjethanandani:
> My one concern here is that there's an operational likely preference that the TZ output should be 'Z' for easier comparison across a multi-geography deployment. However, the YANG type doesn't provide such recommendations.
> 
> @mjethanandani is there any such templated language here recommended by the YANG doctors?

There was an effort to try to define timestamp using 'Z' but it got mired in endless discussion with no agreement in the (netmod) WG. The short answer is, no.

Issue #417, opened 2026-01-23
Add IANA instructions to maintain the IANA registered YANG modules when the associated registries change


jhaas-pfrc:
https://datatracker.ietf.org/doc/html/draft-ietf-netmod-rfc8407bis-28#name-guidance-for-writing-the-ia

`A request to IANA to add a note to the authoritative IANA registry to indicate that any change to the registry must be reflected into the corresponding IANA-maintained module. That is, any changes to the registry must be accompanied by an update to the corresponding IANA-maintained module.`

jhaas-pfrc:
We'll need this for -20.

jhaas-pfrc:
Reviewing iana-bgp-notification, this one cleanly can be documented about maintainence for the BGP notification code and sub-code registry:
https://www.iana.org/assignments/bgp-parameters/bgp-parameters.xhtml#bgp-parameters-3
https://www.iana.org/assignments/bgp-parameters/bgp-parameters.xhtml#bgp-parameters-4

jhaas-pfrc:
iana-bgp-types is overloaded and a broad mix of things:
A cluster of features statements
BGP origin types, which are covered by RFC 4271 but don't currently have a registry. (See https://github.com/ietf-wg-idr/RFC4271bis/issues/102)
AS_PATH segment types (https://github.com/ietf-wg-idr/RFC4271bis/issues/103)
Capabilities, which do have a registry (https://www.iana.org/assignments/capability-codes/capability-codes.xhtml)
Some protocol-specific typedefs.

I'm suspecting the right answer is that we need to split things that are intended to be synched by IANA away from things that aren't.  What's your thought @mjethanandani ?

jhaas-pfrc:
iana-bgp-community-types will be very complicated.  What sort of advice do you envision for IANA for this since our documentation is "add a regex" which is VERY complicated, @mjethanandani ?

jhaas-pfrc:
iana-bgp-rib-types covers things that aren't clearly delineated by specific RFC procedure. The motivation for having this be iana maintained was originally "we need to update this without an RFC".  What's the practice on this now?

suehares:
I got this assigned by Jeff. I'm fine with writing it, but I need to chat with Jeff before working on this part.

jhaas-pfrc:
[...]

PR #490, opened 2026-06-12
Update IANA-maintained YANG modules and split non-IANA content out of iana-bgp-types

mjethanandani:
Address issue #417


draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          IANA-maintained "iana-bgp-notification" YANG module.  The
+          most recent version of the YANG module is available from the
+          "YANG Parameters" registry group.</t>
+
+          <t>IANA is requested to add this note to the
+          "iana-bgp-notification" entry in the "YANG Module Names"
+          registry:</t>
+          <blockquote>New values must not be directly added to the
+          "iana-bgp-notification" YANG module.  They must instead be
+          added to the "BGP Error (Notification) Codes" or "BGP Error
```
jhaas-pfrc: and/or ?

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          Subcodes" registry.</blockquote>
+
+          <t>IANA is requested to add this note to both the "BGP Error
+          (Notification) Codes" and "BGP Error Subcodes"
```
jhaas-pfrc: and/or ?

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          registries:</t>
+          <blockquote>When this registry is modified, the YANG module
```
jhaas-pfrc: either of these registries are

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          "iana-bgp-capabilities"
+          &lt;https://www.iana.org/assignments/iana-bgp-capabilities&gt;
+          must be updated as defined in RFC XXXX.</blockquote>
+
+          <t>When a new Capability Code is added to the "BGP Capability
+          Codes" registry, a new "identity" statement deriving from
+          "bgp-capability" must be added.  The "identity" statement MUST
+          have the substatements defined as described for
+          iana-bgp-notification above.</t>
```
jhaas-pfrc: iana-bgp-capability

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          </blockquote>
+
+          <t>IANA is requested to add this note to both the "Address
+          Family Numbers" and "Subsequent Address Family Identifiers
+          (SAFI)" registries:</t>
+          <blockquote>When this registry is modified, the YANG module
+          "iana-bgp-afi-safi-types"
+          &lt;https://www.iana.org/assignments/iana-bgp-afi-safi-types&gt;
+          must be updated as defined in RFC XXXX.</blockquote>
+
```
jhaas-pfrc, line 903: Here we have a problem.
The AFI registry existed before BGP started leveraging it.  The SAFI registry is particular to BGP.  
We are responsible for all IPv4/IPv6 + SAFI entries.
There are a small number of other AFIs + a small combination of SAFIs that go in there.  

What we need to do is have maintenance done for SAFIs always, and AFIs sometimes.  We need to avoid having IANA go through and say "You've missed all of these possible combinations of AFI+SAFI".

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          IANA-maintained "iana-bgp-rib-types" YANG module.  The most
+          recent version of the YANG module is available from the
+          "YANG Parameters" registry group.</t>
+
+          <t>IANA is requested to create three new registries under a
+          new "BGP YANG RIB Types" registry group:</t>
+
+          <t>Registry 1: BGP Route Ineligibility Reasons</t>
+          <ul>
+            <li>Registration policy: Standards Action</li>
```
jhaas-pfrc, line 982: Ok, so this is the way to trigger and gate things.  That works for me.

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -626,69 +630,337 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+        </section>
+
+        <section title="iana-bgp-rib-types">
+          <t>This document defines the initial version of the
+          IANA-maintained "iana-bgp-rib-types" YANG module.  The most
+          recent version of the YANG module is available from the
+          "YANG Parameters" registry group.</t>
+
+          <t>IANA is requested to create three new registries under a
+          new "BGP YANG RIB Types" registry group:</t>
```
jhaas-pfrc: I see what we're doing here is asking IANA to create registries that didn't exist so the YANG stuff can be maintained.  But these registries aren't "YANG" registries, they're protocol state registries.

@suehares  @keyur-arrcus  I think what these probably should be is "BGP RIB entry state types"?  Other names?

src/yang/example-bgp-configuration-a.1.1.xml
```diff
@@ -36,12 +36,15 @@
               <unreach-decay>500</unreach-decay>
               <keep-history>1000</keep-history>
             </route-flap-damping>
-	    <send-community>bct:standard</send-community>
-	    <send-community>bct:large</send-community>
+	    <send-community>ibct:standard</send-community>
+	    <send-community>ibct:large</send-community>
```
jhaas-pfrc: The indentation of these seems off in the diff review?

src/yang/example-bgp-configuration-a.1.1.xml
```diff
@@ -36,12 +36,15 @@
...
+            <graceful-restart>
+              <enabled>false</enabled>
+            </graceful-restart>
```
jhaas-pfrc: What's motivated the addition of GR

src/yang/iana-bgp-afi-safi-types.yang
```diff
@@ -0,0 +1,155 @@
...
+     described in BCP 14 (RFC 2119) (RFC 8174) when, and only when,
+     they appear in all capitals, as shown here.";
+
+  revision YYYY-MM-DD {
+    description
+      "Initial Version";
+    reference
+      "RFC XXXX: YANG module for Border Gateway Protocol (BGP-4).";
+  }
+
```
jhaas-pfrc, line 65: Shouldn't semver stuff be added in this diff for the modules being updated?

src/yang/iana-bgp-community-types.yang
```diff
@@ -228,288 +131,4 @@ module iana-bgp-community-types {
...
-    reference
-      "RFC 8092: BGP Large Communities Attribute.";
-  }
-
-  /* IPv6 BGP Exteneded Community Types. */
-
-  typedef bgp-ipv6-ext-community-type {
-    type union {
-      type string {
-        pattern 'ipv6\-route\-target:'
```
jhaas-pfrc, line 274: Unfortunately I don't think we can pull out the extended community types.  This is a constantly changing thing as new type/sub-types are added.
If we had a different language way of specifying the named strings or a way to add to the union without revising the whole thing it could move. That's not how it works. :-(

src/yang/iana-bgp-community-types.yang
```diff
@@ -228,288 +131,4 @@ module iana-bgp-community-types {
...
-       is followed by 20 bytes of octet string to support
-       new and experimental type definitions.";
-    reference
-      "RFC 5701: IPv6 Address Specific BGP Extended Community
-       Attribute, Section 3.";
-  }
-
-  /* BGP Extended Community Types. */
-
-  typedef bgp-ext-community-type {
```
jhaas-pfrc, line 354: ditto

jhaas-pfrc (changes requested):

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Ok.

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Ok

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Can you come up with better text?

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Boilerplate suggestion. Feel free to change it.

src/yang/example-bgp-configuration-a.1.1.xml
mjethanandani: I will check.

src/yang/iana-bgp-afi-safi-types.yang
mjethanandani: Ok.

src/yang/iana-bgp-community-types.yang
mjethanandani: There is plenty of this stuff that we'll need to move around. I took a crack at what I remembered.

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: You all are the experts. Does every code have a subcode, in which case it is a 'and', or otherwise it is an 'or'? Or are you suggesting it should be and/or?

draft/draft-ietf-idr-bgp-model.xml
mahesh-arrcus: Actually, the base idenitity is bgp-capability.

draft/draft-ietf-idr-bgp-model.xml
mahesh-arrcus: And is there a AFI registry and YANG model that is in contention?

src/yang/example-bgp-configuration-a.1.1.xml
mahesh-arrcus: Something about the example not verifying unless it is set to 'false'.

draft/draft-ietf-idr-bgp-model.xml
suehares: Mahesh and Jeff: 
IANA denotes it as "Capability codes"
https://www.iana.org/assignments/capability-codes/capability-codes.xhtml

suehares: Please review this point.

draft/draft-ietf-idr-bgp-model.xml
suehares: Suggested text: 
<t> IANA is requested to add this note to the "Address
          Family Numbers"  
          <blockquote>When this registry is modified, the entity 
          registering the new AFI must indicate if the AFI might be passed in BGP. 
          If this is true, the YANG module "iana-bgp-afi-safi-types."
          &lt;https://www.iana.org/assignments/iana-bgp-afi-safi-types&gt;
          must be updated as defined in RFC XXXX. </blockquote>
        </t>

<t>  IANA is requested to add this note to "Subsequent Address Family Identifiers
          (SAFI)" registries:</t>
          <blockquote>When this registry is modified, the YANG module
          "iana-bgp-afi-safi-types"
          &lt;https://www.iana.org/assignments/iana-bgp-afi-safi-types&gt;
          must be updated as defined in RFC XXXX.</blockquote>
        </t> 

how is this text?

draft/draft-ietf-idr-bgp-model.xml
suehares: works for me.

src/yang/example-bgp-configuration-a.1.1.xml
suehares: GR error - yanglint or something else error checked?

src/yang/iana-bgp-afi-safi-types.yang
suehares: Thanks for adding semver Mahesh.

suehares (approved):
I approve all changes.  See my alternate boiler plate for AFI/SAFI as a hint.

src/yang/iana-bgp-community-types.yang
suehares: Extended Communities types need to stay in IANA.  Lots of sub-registries by ext Community type. 
Can we break along sub-registry type?  This needs to be discussed.

src/yang/iana-bgp-community-types.yang
suehares: IANA Extended Communities must be discussed.

draft/draft-ietf-idr-bgp-model.xml
suehares: "and" - is the correct english to indicate both (IMHO).

draft/draft-ietf-idr-bgp-model.xml
suehares: iana-routing-types - has AFI in the Yang module.

draft/draft-ietf-idr-bgp-model.xml
jhaas-pfrc: The flagged point here is you're using notification. It's likely from a prior copy and paste. :-)

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -629,69 +633,353 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+          "iana-bgp-afi-safi-types" entry in the "YANG Module Names"
+          registry:</t>
+          <blockquote>New values must not be directly added to the
+          "iana-bgp-afi-safi-types" YANG module.  They must instead be
+          added to the "Address Family Numbers" / "Subsequent Address
+          Family Identifiers (SAFI)" registries.
+          </blockquote>
+
+          <t>
+	    IANA is requested to add this note to the "Address Family
```
jhaas-pfrc: The indentation has gone off below here.

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -629,69 +633,353 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+	    Numbers" registry:
+	  </t>
+          <blockquote>
+	    When this registry is modified, the entity registering the
+	    new AFI must indicate if the AFI might be passed in
+	    BGP. If this is true, the YANG module
+	    "iana-bgp-afi-safi-types"
+	    &lt;https://www.iana.org/assignments/iana-bgp-afi-safi-types>&gt;
```
jhaas-pfrc: > along with the &gt;?

draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -629,69 +633,353 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
...
+	    When this registry is modified, the YANG module
+	    "iana-bgp-afi-safi-types"
+	    &lt;https://www.iana.org/assignments/iana-bgp-afi-safi-types>&gt;
+	    must be updated as defined in RFC XXXX.
+	  </blockquote>
+
+          <t>When a new AFI/SAFI pair is added, a new "identity"
+          statement deriving from "afi-safi-type" must be added.  The
+          "identity" statement MUST have the substatements defined as
+          described for iana-bgp-notification above.</t>
```
jhaas-pfrc: Another copy/paste error - this isn't for notificatoin?

src/yang/example-bgp-configuration-a.1.1.xml
jhaas-pfrc: Ok, the validation requirement here would be wrong.  I'll see if I can debug it.

jhaas-pfrc (changes requested):

src/yang/example-bgp-configuration-a.1.1.xml
jhaas-pfrc: I've locally removed the graceful restart from the examples and am not seeing validation issues.  Can you paste me your output, @mjethanandani ?

mjethanandani: For some reason, I am unable to reproduce the issue. Undoing the changes.

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Let me know if you are still seeing the indentation off. I have tried to replace all tabs with spaces.

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Fixed

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: They have AFIs as enums!! 

Do we need to remove all AFI definitions in our YANG module and the IANA instructions for AFI in that case?

src/yang/iana-bgp-community-types.yang
mjethanandani: Let me know if I have put back what we need into the IANA module.

draft/draft-ietf-idr-bgp-model.xml
jhaas-pfrc: I'm a bit unclear what the complaint in these last two comments is referring to?

src/yang/iana-bgp-community-types.yang
jhaas-pfrc: I believe the structure and split for communities is now correct.

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Here is the issue we are dealing with.

RFC 8294[https://datatracker.ietf.org/doc/html/rfc8294#page-38](url) has a YANG module that defines something called `address-family` and something called `bgp-safi`, which are typedefs with an enum definition for each of AFI and SAFI. We should be using those, rather than defining our own, since our iana-afi-safi-types YANG module overlaps with that YANG module.

I would have ideally defined them as identities, but we have what we have, and now we have to deal with it.

mjethanandani: As specified, users will have to specify AFI and SAFI as two separate enums, instead of as a tuple. BTW, OpenConfig also defines it as we do. To keep our model, we would have to override the definition in iana-routing-types.

Let me know how you would like to proceed.

jhaas-pfrc: I understand the complaint now.
You'll see we use the existing iana maintained afi and safi entries selectively in our own module.  However, we prefer to identity that is the combined type.  It'd be interesting to see what the forensic history of this choice is vs. when the IANA module came into being, but only academically.

For those types, enum is the way to go.  In circumstances where you are given an on-the-wire output that doesn't map to something you don't know, the enum would at least let you see what it was while the typedef wouldn't have.

The useful property of a BGP-specific combination identity is it also implies a level of "support" for that combination.  You'll recognize this from the question about how to write the text for maintaining the identity based on maintenance of the afi.  Not all combinations of afi+safi will be useful for bgp.  Even for a given family, not all combinations of safi + afi are guaranteed to be useful.  And perhaps we need to write our text such IANA seeks expert advice when going to register a new afi or safi before adding it to the iana-maintained bgp identity?

mjethanandani: If I understand it correctly, you are suggesting we go with using the AFI/SAFI definitions in the ietf-routing-types module and adjust our model to that.

Let me send an update to accomodate that suggestion.

jhaas-pfrc: I'm suggesting we keep what we have.
This means that the afi and safi registries have impacts in two different sets of IANA maintained types: The afi and the safi enums. The afi-safi identities we're creating.

mjethanandani: Hmm! I wonder how IANA (and IESG) will react to having to maintain two different YANG modules for the same IANA registry. I will go ahead with this suggestion.

But I still need your approval on the PR, unless there are more changes you are looking for.

suehares (approved):

draft/draft-ietf-idr-bgp-model.xml
suehares: I'm fine with Jeff's approach

suehares (approved):
I approve what I see here.  I'm fine with Jeff's approach to afi-safi.

PR #497, opened 2026-07-13
Address Maria's feedback on -20

### LINKED ISSUE #496: Issues flagged by Maria in -20
```
Sec. 7.1 feature damping {
description
"Weighted route dampening is supported.";
}
```

This should be damping.

```
Sec. 7.3.1 leaf identifier {
type yang:dotted-quad;
0description
"BGP Identifier of the router - an unsigned 32-bit,
non-zero integer that should be unique within an AS.
The value of the BGP Identifier for a BGP speaker is
determined upon startup and is the same for every local
interface and BGP peer.";
reference
"RFC 6286: AS-Wide Unique BGP ID for BGP-4. Section 2.1";
}
What about an analogous type to rr-cluster-id-type, as RFC 6286 specifies it as uint32?
```

The request here is to consider using a similar typedef for BGP-ID.  While semantically valid, I'm unaware of implementations that don't use dotted quad for this.  Audit.

```
      container neighbors {
        description
          "Configuration for BGP neighbors.";

```

The complaint here is this isn't *only* configuration.  Fix the description.

```

leaf local-restarting {
type boolean;
config false;
description
"This flag indicates whether the local neighbor is
currently restarting. The flag is cleared after all
NLRI have been advertised to the peer, and the
End-of-RIB (EOR) marker has been cleared.";
}
Does this mean “the local node / session / connection is restarting?” For me, a neighbor is not the local node. RFC 4724 uses the term Speaker. Also just several items below, inleaf session-state, the term Neighbor is used for the remote one only.
```

Audit vs. "local BGP speaker" terminology.

```
action clear {
if-feature "ibt:clear-statistics";
description
"Clear statistics action command.
Execution of this command should result in all the
counters to be cleared and set to 0.";
I’m not completely sure but do I interpret correctly that container queues, which are gauge32, are not counter, and therefore not cleared? This may deserve a clarification, what exactly is cleared.
```

Update terminology to consistently refer to various statistics keeping leaf nodes rather than name by type.

```
container route-flap-damping
This deserves an explicit reference to RFC 2439.
```

Add references.

```
leaf-list send-community {
if-feature "ibct:send-communities";
type identityref {
base "ibct:send-community-feature";
}
description
"When supported, this tells the router to propagate any
prefixes that are attached to these community-types.";
}
This reads incomprehensible to me. How is a prefix attached to a community type?
```

Improperly documented. This should refer only to propagation of the communities and not a route-filter.

```
container dynamic-peers {
list dynamic-peer-list {
key "prefix";
While one may intuitively understand how this works, I would, again, like to see a more detailed explanation of the semantics of dynamic peers. I know that it’s quite a common feature but it’s not specified formally and there is yet again a risk of semantic confusion between implementations.
```

Consider what we would do to document this feature in the draft since there's no protocol RFC reference.

jhaas-pfrc:
Closes #496.

Fix warning from long line
Update description for container neighbors to note operational state
Address graceful restart local naming issues
Update text on 'clear bgp statistics' to be clear that it is statistic related, not data type.
Add reference for RFC 2439 to container route-flap-damping
Fix description for send-community
Add text covering dynamic-peering sessions

Note that this does not address a more flexible type for leaf identifier to permit uint32 as a union.

### DIFF
--- a/draft/draft-ietf-idr-bgp-model.xml
+++ b/draft/draft-ietf-idr-bgp-model.xml
@@ -320,6 +320,12 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-sub-tree.txt)
         neighbor-specific configuration options which are relevant to the
         entire BGP instance.</t>
 
+        <t>In addition to explicit neighbor configuration, remote BGP neighbors
+        may establish sessions dynamically with the local BGP speaker.  This is
+        configured using the 'dynamic-peers' container which uses a prefix-list
+        to control IP subnets permitted to establish such dynamic peering
+        sessions.</t>
+
         <t>The model makes the simplifying assumption that most of the
         configuration items are available at all levels of the hierarchy. That
         is, very little configuration is specific to a particular level in the
--- a/src/yang/ietf-bgp-common.yang
+++ b/src/yang/ietf-bgp-common.yang
@@ -363,6 +363,8 @@ submodule ietf-bgp-common {
       description
         "Routes learned via BGP are subject to weighted route
          damping.";
+      reference
+        "RFC 2439: BGP Route Flap Damping.";
     }
     leaf-list send-community {
       if-feature "ibct:send-communities";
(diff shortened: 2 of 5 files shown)


mjethanandani (approved):

suehares (approved):
changes ares ok with me.

PR #501, opened 2026-08-07
Issue 500: Session security needs if-features

### LINKED ISSUE #500: Authentications should require an if-feature?
Raised as part of a BMP YANG model review at IETF-126 GROW meeting, the BGP model likely requires an if-feature.  otherwise, a non-compliant implementation is forced to do a deviation to remove the options.

jhaas-pfrc:


draft/draft-ietf-idr-bgp-model.xml
```diff
@@ -616,6 +616,7 @@ INSERT_TEXT_FROM_FILE(../bin/ietf-bgp@YYYY-MM-DD-rib-tree.txt)
             URI:urn:ietf:params:xml:ns:yang:iana-bgp-capabilities
             URI:urn:ietf:params:xml:ns:yang:iana-bgp-notification
             URI:urn:ietf:params:xml:ns:yang:iana-bgp-rib-types
+            URI:urn:ietf:params:xml:ns:yang:iana-bgp-session-security
```
mjethanandani, line 619: I believe there is some text at the beginning of the IANA section that talks about the number of registrations.

src/yang/ietf-bgp-session-security.yang
```diff
@@ -0,0 +1,167 @@
...
+    "WG Web:   <http://datatracker.ietf.org/wg/idr>
+     WG List:  <idr@ietf.org>
+
+     Authors: Mahesh Jethanandani (mjethanandani at gmail.com),
+              Keyur Patel (keyur at arrcus.com),
+              Susan Hares (shares at ndzh.com),
+              Jeffrey Haas (jeffrey.haas at hpe.com).";
+
+  description
+    "This sub-module contains groupings particular to BGP session
```
mjethanandani, line 28: sub-module in YANG has a special meaning, which this is not. Just call it a module.

src/yang/ietf-bgp-session-security.yang
```diff
@@ -0,0 +1,167 @@
...
+  feature bgp-session-security-tcp-ao {
+    description
+      "Support for the TCP Authentication Option for neighbors.";
+    reference
+      "RFC 5925: The TCP Authentication Option.";
+  }
+
+  feature bgp-session-security-tcp-md5 {
+    description
+      "Support for TCP-MD5 for neighbors.";
```
mjethanandani, line 72: Drop in the sentence from the draft that says MD5 is being supported for legacy reasons.

draft/draft-ietf-idr-bgp-model.xml
suehares: I'm confused by Mahesh's comment.  Is there something that needs to change?

src/yang/ietf-bgp-session-security.yang
suehares: Mahesh - just for my information what makes a sub-module a special meaning?

src/yang/ietf-bgp-session-security.yang
suehares: Mahesh - will we get into trouble with Security by dropping the TCP-MD5 is legacy?

suehares (approved):
I'm fine with the changes.  However, I am concern that security area will complain about TCP-MD5.  

Sue

draft/draft-ietf-idr-bgp-model.xml
mjethanandani: Section 5 of the draft says:

This document registers nine URIs, twelve IETF YANG modules and submodules, and five IANA-maintained YANG modules.

We now adding another module. So number of URIs jumps to ten, number of IETF YANG moules and submodules is thirteen.

src/yang/ietf-bgp-session-security.yang
mjethanandani: https://www.rfc-editor.org/rfc/rfc7950.html#section-4.2.1

src/yang/ietf-bgp-session-security.yang
mjethanandani: I said "Drop in", not drop out the reference to TCP-MD5.

src/yang/ietf-bgp-session-security.yang
jhaas-pfrc: Paste error. Corrected.

src/yang/ietf-bgp-session-security.yang
jhaas-pfrc: Updated in current diff.

jhaas-pfrc:
> I'm fine with the changes. However, I am concern that security area will complain about TCP-MD5.

Of *course* the security ADs will complain.  But similarly, we can't get rid of it.  So, as usual, we argue.

draft/draft-ietf-idr-bgp-model.xml
jhaas-pfrc: Updated in diff.

mjethanandani (approved):

